PR #1, opened 2020-01-23
Initial commit of openmrs-config-pihemr

mogoodrich:


mseaton:
This is now merged, but a few quick comments:
* I think we agreed to put this under "org.pih.openmrs", not "org.pih".
* When you merge something like this, would be good to squash when you do.  Keeps the commit history cleaner.

PR #33, opened 2020-05-07
UHM-4439: Migrate GlobalPropertyBundle out of code and into config (WIP)

mogoodrich:


brandones:
I notice a lot of things being factored out into `constants.yml` that are only used once.

```xml
        <globalProperty>
            <property>htmlformentry.timeFormat</property>
            <value>${javaTimeFormat}</value>
        </globalProperty>
```

`${javaTimeFormat}` is only used once. Please don't make me go look it up.

Similarly, rather than

```xml
        <globalProperty>
            <property>labworkflowowa.didNotPerformReason</property>
            <value>${concept.didNotPerformReason.uuid}</value>
        </globalProperty>
```

and off in another file

```yaml
  didNotPerformReason:
    uuid: "5dc35a2a-228c-41d0-ae19-5b1e23618eda"
```

What I would much rather work with would be

```xml
        <globalProperty>
            <property>labworkflowowa.didNotPerformReason</property>
            <!-- CIEL:165182 Reason lab procedure not performed -->
            <value>5dc35a2a-228c-41d0-ae19-5b1e23618eda</value>
        </globalProperty>
```

Then I don't have to be cross-referencing mapping names to get UUIDs, nor inferring concept FSNs from those mapping names -- all the information I need is right there. If I need to use that concept again elsewhere, I'll grep for it and factor it out when needed.

This is just my preference, based on my experience working with all the other places where we've factored out UUIDs into only-slightly-less-opaque-if-not-misleading constants.

Sorry for the length, but I imagine this will generate some controversy.

mogoodrich:
Let me ponder this... it's a fair point that just using the uuid and adding a nice comment like "CIEL:165182 Reason..." may be a better approach.  I will say that as for "... If I need to use that concept again elsewhere, I'll grep for it and factor it out when needed..." people generally won't do that unless we are good about setting up concepts in a constants file to start with.

mogoodrich:
So this is a WIP as well, but this was exactly the kind of feedback I wanted at this point... :)

mseaton:
I also understand where you are coming from @brandones .  There are certainly settings that we can likely leave as hard-coded in to global properties and be happy with it.  I probably wouldn't put concept uuids in that list, though we could debate about it.

I actually lean (probably unsurprisingly) to using the concepts.yml file by default and referring to it in the global properties files for much _more_, and I agree with Mark that we are trying to set up some best practices that this will encourage.  People will get in the habit of looking in the constants.yml file for whatever they are looking for first, rather than going to the gp.xml file first.  And I'd rather not have to make judgement calls and have debates in every PR about where things should go.  If we set up the standard to use constants and follow it regardless if there is active need, then this will be easier to maintain and we won't constantly be asking or debating where the value should be defined.

To me it isn't just about reuse of a constant across multiple points, but is also about a single point of reference for as much of our settings as possible.  Looking at this one file will give a lot of insight into what settings exist in the system, and having things organized by name/domain will allow for us to see things more clearly than the GP xml alone, which has a hodgepodge of global property names following no discernible format or convention.  I'd rather get in the habit of looking in constants.xml for something first - rather than having to grep the global property xml and a dozen other Iniz files that are written for specific domains.

One thing that I've always thought would be interesting to see as well, would be if we can _generate_ the global properties.yml from our constants.yml.  Not sure how easy that would be to do, and it probably wouldn't add any real readability or utility, so probably not worth it, but something worth highlighting.

brandones:
Every item factored out into `constants.yml` introduces a naming problem. Most of the time, people have tended to name constants after the first place that they happened to be used, independently of what they actually refer to. So I ask that even if we are going to create a new hodgepodge of names, that we establish very strong conventions. Perhaps we can start with "concept constant names should be the FSN in camelCase."

But then, of course, the mapping from global property key to concept constant name is no longer obvious, and one can't just look at the constants file and infer how the global properties are set. Highlighting the indirection problem.

mseaton:
It's a good point @brandones .  Naming is hard and it's true that this could lead to challenges.  One of the things I've been trying to advocate for is a "key" for all metadata that we can refer to uniquely to look up items that is independent of uuids or translations and display names.  Concept mapping (and metadata mapping codes more generally) are the closest thing we have to this, and I don't see us adopting "keys" any time soon.  So I'd probably advocate for our constant names having a 1:1 relationship with concept/metadata mapping codes of a certain type.

But ultimately it is not that many concepts that we will be referring to in constants explicitly - only those that have particular business meaning within our PIH EMR application.  It's not like we need a constant for every clinical question that is asked or their coded answers.  For those that we do explicitly code around, having a constant name that these refer to seems reasonable and not like a huge burden.

mogoodrich:
This is a key line from @mseaton "To me it isn't just about reuse of a constant across multiple points, but is also about a single point of reference for as much of our settings as possible. "  Extracting out into constants.yml isn't mean to make reading gp.xml (nor does it!)...

However, I am pretty swayed by @brandones's argument reading naming conventions and indirections.  What if we standardize on the following naming convention for concept mappings:

concept.cielMappingName or concept.pihMappingName, preferring CIEL if present, and if not falling back to PIH.  And if any concept doesn't have a meaningful mapping name in CIEL or PIH we *don't* make it into a constant?

For other metadata we just use the name...

I *do* also want to keep javaTimeFormat: "h:mm aa" has a constant, because this really is defining a standard more than anything else.

Thoughts?

Not sure when I'm going to get back to this, but I could do an updated draft PR with the more format "keys" and see what we think.

brandones:
> concept.cielMappingName or concept.pihMappingName, preferring CIEL if present, and if not falling back to PIH. And if any concept doesn't have a meaningful mapping name in CIEL or PIH we _don't_ make it into a constant?

I think CIEL codes are really useful when building things that other organizations might want to use. I don't think this is such a case, so I'd really prefer to see a human-readable name rather than a big integer.

> I _do_ also want to keep javaTimeFormat: "h:mm aa" has a constant, because this really is defining a standard more than anything else.

It's already a standard by virtue of being the unique place that a time format to be used throughout the application is defined.

As an aside, I'm surprised our African colleagues haven't complained about [AM/PM](https://en.wikipedia.org/wiki/Date_and_time_representation_by_country#/media/File:12_24_Hours_World_Map.svg).

mogoodrich:
Oh, duh @brandones ... so, right, all CIEL codes are numbers by definition.  :)  PIH mapping can either be numbers or names (just the FSN at the time of mapping iirc) so when I said " meaningful mapping name in CIEL or PIH we don't make it into a constant" I meant "human-readable".  So revised proposal would be to use the PIH mapping if there's a human-readable non-numberic one, otherwise potentially not make it into a constant.

If the majority of the concepts we are using in GPs don't have particularly readable mappings right now I'm inclined to bag the whole concept as constants stuff for now.

RE: time "javaTimeFormat", that's actually the point... we are defining "htmlformentry.timeFormat" which *only* sets the time format within the HTML Form Entry module.  We could extract it out into a constant at a later time if we end up needing to use it in another place *if* whoever added it took the time to look through all the global properties to see if there was a similar property.

mseaton:
At the end of the day, we should just use the same process and judgement here as we have within the Java codebase.  At times, we have create constants to refer to settings in a central place and then used these constants.  At other times, we have hard-coded values in place and not created them as constants.  We don't have to be dogmatic and, ultimately, will not likely have a lot of downstream control over how our implementations decide to iterate and evolve upon these.

I 1000% percent agree that the more best practices and patterns we can establish now the better, if they will lead downstream users to follow good practices going forward.  That's the crux of the arguments here.  I don't think we should be dogmatic, but if we can create some patterns to follow using constants so that others can see how it is done and follow it, that's a good thing.

mogoodrich:
Okay, so I took another pass at this, and ended up *only* putting general, reusable concepts as constants (ended up only three of them so far).  But I did start the practice of making sure we had a comment with the name of the concept for each concept that we refer to directly by uuid in gp.xml.

I added some comments to the constants.yml file as well...

Generally I'm happy with the way things worked out... still a work-in-progress, as I have to start moving GPs from other files over as well...

brandones:
If a site wants to override one of these GPs, can they simply define it in their own gp.xml file?

mogoodrich:
Unfortunately not as things currently work @brandones ... if a site defines a "gp.xml" file in it's config, it will override the other "gp.xml" entirely, so I had to create a naming convention like "gp_zl.xml".

If Iniz has a way to control the order of loading of files, then if we set "gp_zl.xml" to load second, that would be a way of achieving this.

But, otherwise, if a site wants to customize something we will just need to extract that GP out of the main PIH EMR and into all the site GP files, even if it's identical for all but one site.

brandones:
[Asked in Slack](https://openmrs.slack.com/archives/CPC20CBFH/p1589408142020500)

mseaton:
> But, otherwise, if a site wants to customize something we will just need to extract that GP out of the main PIH EMR and into all the site GP files, even if it's identical for all but one site.

Wouldn't this be a situation where we'd be better served setting the GP value to a config constant, setting this config constant to a default value in config-pihemr, and then only overriding it in the config-xyz projects that need to?

brandones:
Maybe, but then we're introducing downstream dependencies of pihemr GP declarations and how they relate to their constants. Then we have to worry about breaking those dependent configs.

Also, then if I need to make a change to a GP downstream, I need to look at the pihemr GP file, cross-reference the constants file, then search the pihemr GP file and my own GP file to see what other references there are to that constant, and then decide to either 1. override that constant or 2. override the GP because overriding the constant would cause unintended side-effects. It is likely that people would forget some of these steps frequently.

@mogoodrich , looking at the PR as it is now, I really like the balance you've struck. I would be very happy working with these files.

mogoodrich:
@brandones  I think @mseaton  was referring to "config-xyz" files, *not* the constants.yml.  I had forgotten until I was finishing the second part of this ticket, but they also provide a means to set GPs and to customize them by country or site:

https://github.com/PIH/openmrs-config-zl/blob/master/configuration/pih/pih-config-mirebalais.json#L8

I'm actually leaving the few GPs that need to differ between sites in the same country in these configs, because we don't have that level of granuality in the Iniz config yet (not because of any limitation of Iniz but because we don't yet support making different "configuration" directories for two sites in the same country).

So there are several different approaches we could take here... and can figure out which one to take when the time comes... :)

brandones:
Oh, interesting! Good to know about that, thank you.

mseaton:
No, I meant the constants.yml :)

mogoodrich:
Ah, okay @mseaton ... well, I'm not going to reopen the debate until the we run into that scenario... :)

PR #34, opened 2020-05-19
Remove English messages from ht, tx push -t -l ht, tx pull

brandones:


brandones:
Okay, after pushing and pulling from transifex, there's no change. Closing.

PR #38, opened 2020-06-03
UHM-4608 EncounterTypeConfig should not reference constants library

brandones:
https://pihemr.atlassian.net/browse/UHM-4608


mogoodrich:
@brandones  looks good to me, as long as you double-checked the constants when you copied them... :)

@Inball we need this because we pulled the encounter config out of code and into config and it was still references constants in code.  This was my mistake, but it cost me a couple hours of cleanup yesterday because it completed hosed Sierra Leone, just want to make sure if doesn't happen when we roll out elsewhere...

brandones:
Awesome, thanks @mogoodrich !

brandones:
@lnball Yeah more just keeping you abreast of changes in this code, since you help maintain these config repos.

PR #40, opened 2020-06-04
Fix background color of labResults form on tablet

brandones:
https://pihemr.atlassian.net/browse/UHM-4579

Screenshot from 2020-06-04 14-55-33


brandones:
Trivial; merging

PR #42, opened 2020-06-18
Uhm 4587 encounter exports covid

jmbabazi:


configuration/pih/reports/sql/covid19encounterexport.sql
```diff
@@ -6,98 +6,1328 @@
...
+    FROM
+        person_attribute a
+            INNER JOIN
+        person_attribute_type t ON a.person_attribute_type_id = t.person_attribute_type_id
+    
+    WHERE
+        a.value = 'true'
+        AND t.name = 'Test Patient');
+
+UPDATE temp_encounter
```
mseaton, line 208: I personally much prefer the formatting where these update statements are on a single line.  If find it _much_ easier to read.

configuration/pih/reports/sql/covid19encounterexport.sql
```diff
@@ -6,98 +6,1328 @@
...
+UPDATE temp_encounter 
+SET 
+    health_care_worker = OBS_VALUE_CODED_LIST(encounter_id, 'CIEL', '5619', 'fr');
+
+UPDATE temp_encounter 
+SET 
+    pregnant = OBS_VALUE_CODED_LIST(encounter_id, 'CIEL', '5272', 'fr');
+
+-- last menstruation date
+UPDATE temp_encounter te
```
mseaton, line 237: If we have to do a join here, we are missing a function.  Why are we doing this instead of adding the appropriate function to retrieve an obs value datetime?

configuration/pih/reports/sql/covid19encounterexport.sql
```diff
@@ -6,98 +6,1328 @@
...
+        LEFT JOIN
+    obs o ON o.concept_id = CONCEPT_FROM_MAPPING('CIEL', '1427')
+        AND o.person_id = te.patient_id
+        AND o.encounter_id = te.encounter_id
+        AND o.voided = 0 
+SET 
+    te.last_menstruation_date = o.value_datetime;
+
+-- estimated delivery date
```
mseaton, line 246: Same comment as above - use a function to get the datetime value.

configuration/pih/reports/sql/covid19encounterexport.sql
```diff
@@ -6,98 +6,1328 @@
...
+            'CIEL',
+            '129317');
+
+-- outcome
+UPDATE temp_encounter 
+SET 
+    outcome = OBS_VALUE_CODED_LIST(encounter_id, 'CIEL', '161033', 'fr');
+
+-- date of delivery
+UPDATE temp_encounter te
```
mseaton, line 271: Same comment as above - use a function to get the datetime value.

configuration/pih/reports/sql/covid19encounterexport.sql
```diff
@@ -6,98 +6,1328 @@
...
+-- home medication
+UPDATE temp_encounter 
+SET 
+    home_medications = OBS_VALUE_TEXT(encounter_id, 'CIEL', '162165');
+
+UPDATE temp_encounter 
+SET 
+    allergies = OBS_VALUE_TEXT(encounter_id, 'CIEL', '162141');
+
+UPDATE temp_encounter te
```
mseaton, line 309: Same comment as above - use a function to get the datetime value.

configuration/pih/reports/sql/covid19encounterexport.sql
```diff
@@ -6,98 +6,1328 @@
...
+SET 
+    comorbidities = OBS_VALUE_CODED_LIST(encounter_id, 'PIH', '12976', 'fr');
+
+-- comorbidities
+UPDATE temp_encounter 
+SET 
+    available_comorbidities = OBS_VALUE_CODED_LIST(encounter_id, 'CIEL', '162747', 'fr');
+
+-- other comorbidities
+UPDATE temp_encounter te
```
mseaton, line 338: We use the obs comments field?  OK.  If so, please move this into a function and then use the function.

configuration/pih/reports/sql/covid19encounterexport.sql
```diff
@@ -6,98 +6,1328 @@
...
+UPDATE temp_encounter 
+SET 
+    bp_diastolic = OBS_VALUE_NUMERIC(encounter_id, 'CIEL', '5086');
+
+UPDATE temp_encounter 
+SET 
+    SpO2 = OBS_VALUE_NUMERIC(encounter_id, 'CIEL', '5092');
+
+-- room air
+UPDATE temp_encounter
```
mseaton, line 403: Is it more readable to have this statement on 7 lines of code?  Can't we just make all of these 1 line of code?

configuration/pih/reports/sql/covid19encounterexport.sql
```diff
@@ -6,98 +6,1328 @@
...
+UPDATE temp_encounter 
+SET 
+    o2therapy = OBS_VALUE_NUMERIC(encounter_id, 'CIEL', '165986');
+
+-- analgesic comments/description
+UPDATE temp_encounter 
+SET 
+    analgesic_specified = OBS_VALUE_TEXT(encounter_id, 'CIEL', '163206');
+
+##### IVF
```
mseaton, line 566: Hard for me to totally understand what is going on below in this IVF section with these temporary tables.  Might need more discussion.

configuration/pih/reports/sql/covid19encounterexport.sql
```diff
@@ -6,98 +6,1328 @@
...
+#######Radiology
+UPDATE temp_encounter 
+SET 
+    chest_xray = OBS_SINGLE_VALUE_CODED(encounter_id,
+            'PIH',
+            'Radiology procedure performed',
+            'CIEL',
+            '165152');
+
+UPDATE temp_encounter e
```
mseaton, line 841: Whatever we are doing here, we need to extract it out into a function.

configuration/pih/reports/sql/covid19encounterexport.sql
```diff
@@ -6,98 +6,1328 @@
...
+
+UPDATE temp_encounter 
+SET 
+    cardioUS = OBS_SINGLE_VALUE_CODED(encounter_id,
+            'PIH',
+            'Radiology procedure performed',
+            'CIEL',
+            '163041');
+
+UPDATE temp_encounter e
```
mseaton, line 876: Same here.  This complicated update, run multiple times, needs to be pulled into a function.

configuration/pih/reports/sql/covid19encounterexport.sql
```diff
@@ -6,98 +6,1328 @@
...
+    cardioUS_findings = o1.value_text;
+
+UPDATE temp_encounter 
+SET 
+    abUS = OBS_SINGLE_VALUE_CODED(encounter_id,
+            'PIH',
+            'Radiology procedure performed',
+            'CIEL',
+            '845');
+UPDATE temp_encounter e
```
mseaton, line 910: Same comment as above

configuration/pih/reports/sql/covid19encounterexport.sql
```diff
@@ -6,98 +6,1328 @@
...
+                    WHERE
+                        encounter_type = @covid_admission_encounter_type))
+            AND concept_id = CONCEPT_FROM_MAPPING('PIH', 'Radiology report comments')
+            AND voided = 0) o1 ON e.patient_id = o1.person_id
+        AND e.encounter_id = o1.encounter_id 
+SET 
+    abUS_findings = o1.value_text;
+
+-- Other finding
+UPDATE temp_encounter e
```
mseaton, line 938: Same comment as above

configuration/pih/reports/sql/covid19encounterexport.sql
```diff
@@ -6,98 +6,1328 @@
...
+                        encounter
+                    WHERE
+                        encounter_type = @covid_admission_encounter_type))
+            AND concept_id = CONCEPT_FROM_MAPPING('PIH', 'Radiology procedure performed')
+            AND voided = 0) o1 ON e.patient_id = o1.person_id
+        AND e.encounter_id = o1.encounter_id 
+SET 
+    radiology_other = CONCEPT_NAME(o1.value_coded, 'fr');
+
+UPDATE temp_encounter e
```
mseaton, line 965: Same comment as above

configuration/pih/reports/sql/covid19encounterexport.sql
```diff
@@ -6,98 +6,1328 @@
...
+
+UPDATE temp_encounter 
+SET 
+    disposition = OBS_VALUE_CODED_LIST(encounter_id,
+            'PIH',
+            'Hum Disposition categories',
+            'fr');
+UPDATE temp_encounter 
+SET 
+    admission_ward = (SELECT
```
mseaton, line 1000: If we don't already have a function to get a location_id from a lookup (name, uuid, metadata mapping), then it should be added and utilized here.

configuration/pih/reports/sql/covid19encounterexport.sql
```diff
@@ -6,98 +6,1328 @@
...
+-- overall_condition
+update temp_encounter set overall_condition = obs_value_coded_list(
+	encounter_id,
+    'CIEL',
+    '159640',
+    'fr'
+);
+
+-- new symptom
+UPDATE temp_encounter te LEFT JOIN
```
mseaton, line 1046: We're using functions well in all these below, but there must be a way to simplify this use case into a more concise function.

PR #44, opened 2020-07-13
UHM-4691: Enable SQL functions to be defined in the configuration dir…

mogoodrich:
…ectory, by SQL files


configuration/pih/reports/sql/procedures/liquibase.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd">
+
+    <changeSet id="20200713-pihcore-create-functions" author="mgoodrich" runAlways="true" runOnChange="true">
+        <sqlFile endDelimiter="#" path="create_functions.sql" relativeToChangelogFile="true" stripComments="true" />
```
mogoodrich, line 9: Note the "relativeToChangelogFile" parameter, so that it looks for this file relative to the directory that the changelog file lives in.

I had also to update the schema references in the databaseChangeLog to 2.0 in order for it not to throw an error on the relativeToChangelog tag (we are using liquibase 2.0 in Core 2.3)

PR #45, opened 2020-07-15
Uhm 4680 covid admission

jmbabazi:


jmbabazi:
@mseaton, @mogoodrich and @dmdesimone . Creating this PR. Would you please review when you get a chance.

configuration/pih/petl/jobs/covid19/admission/job.yml
```diff
@@ -0,0 +1,13 @@
...
+    datasource: "mysql/openmrs.yml"
+    query:  "covid19/admission/source.sql"
+
+  load:
+    datasource: "sqlserver/openmrs_extractions.yml"
+    table: "covid_admission"
+    schema: "covid19/admission/target.sql"
+
+schedule:
+  cron: "30 */4 * ? * *"
```
mogoodrich, line 13: I'm thinking it makes sense to run these once a day, early in the morning, after the Mirebalais Reporting server has "refreshed" itself with the Mirebalais Production database? I set the vaccincation e job to run at 5am based on when it looked like the percona stuff ran, but correct me if I'm wrong @jmbabazi .

The other important note was that the cron format here has a seconds component... this confused me a bit in the week.  So, for instance, the above cron says "run every 4 minutes at 30 seconds past the minute", not "run every 4 hours, at 30 minutes past the hour"

PR #46, opened 2020-07-21
Create functions for SL dispensing export

dmdesimone:
Functions needed for SL-176, which is a subtask of SL-175, the dispensing report for Sierra Leone

### DIFF
--- a/configuration/pih/liquibase/sql/create_functions.sql
+++ b/configuration/pih/liquibase/sql/create_functions.sql
@@ -1090,3 +1090,180 @@ BEGIN
 END
 
 #
+
+-- This function accepts visit_id, mapping source, mapping code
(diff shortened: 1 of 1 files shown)


configuration/pih/liquibase/sql/create_functions.sql
```diff
@@ -1090,3 +1090,180 @@ BEGIN
...
+    DETERMINISTIC
+
+BEGIN
+
+  DECLARE enc_name_out varchar(255);
+
+  select et.name into enc_name_out
+  from encounter_type et
+  where et.retired = 0
+  and (et.encounter_type_id = _type_or_uuid or et.uuid = _type_or_uuid)
```
mseaton, line 1165: If you tested this and it works, fine, but obviously the data type is different

PR #47, opened 2020-07-22
UHM-4685: Define a configuration file format and mechanism for loadin…

mogoodrich:
…g reports from the configuration directory


mogoodrich:
Makes sense re: the relative directory for the sql file...

PR #50, opened 2020-08-17
Add return visit date

dmdesimone:
Add return visit date for discharged COVID-19 patients for UHM-4721


jmbabazi:
This looks great@ thanks @dmdesimone

PR #58, opened 2020-09-10
HIV viral load report

jmbabazi:


configuration/pih/petl/jobs/hiv/viral_load/source.sql
```diff
@@ -0,0 +1,126 @@
...
+-- hiv constructs table
+CREATE TEMPORARY TABLE temp_hiv_constructs
+(
+patient_id 				INT,
+encounter_id 			INT
+);
+
+INSERT INTO temp_hiv_constructs (patient_id, encounter_id)
+SELECT person_id, encounter_id FROM obs WHERE voided = 0 AND concept_id = CONCEPT_FROM_MAPPING("PIH", "HIV viral load construct") AND encounter_id IN (SELECT encounter_id FROM encounter
+WHERE encounter_type =(SELECT encounter_type_id FROM encounter_type WHERE name = "Laboratory Results"));
```
mogoodrich, line 17: Right now I think this is the only encounter type that has VL results, but is there any harm in removing this clause and just searching on the "HIV viral load construct" just in case this changes in the future?

configuration/pih/petl/jobs/hiv/viral_load/source.sql
```diff
@@ -0,0 +1,126 @@
...
+DROP TABLE IF EXISTS temp_days_since_vl;
+
+-- hiv constructs table
+CREATE TEMPORARY TABLE temp_hiv_constructs
+(
+patient_id 				INT,
+encounter_id 			INT
+);
+
+INSERT INTO temp_hiv_constructs (patient_id, encounter_id)
```
mogoodrich, line 15: A nitpick, but would this be better as somethign like "temp_hiv_construct_encounters"?

configuration/pih/petl/jobs/hiv/viral_load/source.sql
```diff
@@ -0,0 +1,126 @@
...
+FROM temp_viral_load tvl
+-- speciment collection date 
+JOIN temp_vl_sample_taken_date tvld ON tvld.encounter_id = tvl.encounter_id 
+-- days since last viral load
+LEFT JOIN temp_days_since_vl tdl ON tvl.patient_id = tdl.patient_id AND tdl.result_date = tvl.vl_result_date
+-- index ascending
+LEFT JOIN temp_index_asc ON tvl.encounter_id = temp_index_asc.encounter_id
+-- index descending
+LEFT JOIN temp_index_desc ON tvl.encounter_id = temp_index_desc.encounter_id
+ORDER BY tvl.patient_id;
```
mogoodrich, line 126: There's an assumption here that's there's only only VL construct per encounter, which is totally fine to make, but we should note that assumption in a comment at the top this file, I think.

configuration/pih/petl/jobs/hiv/viral_load/source.sql
```diff
@@ -0,0 +1,126 @@
...
+
+-- specimen cpllection date table
+CREATE TEMPORARY TABLE temp_vl_sample_taken_date
+(
+encounter_id 			INT,
+vl_sample_taken_date	DATE
+);
+
+INSERT INTO temp_vl_sample_taken_date(encounter_id, vl_sample_taken_date)
+SELECT encounter_id, DATE(encounter_datetime) FROM encounter WHERE encounter_id IN (SELECT encounter_id FROM temp_hiv_constructs);
```
mogoodrich, line 27: Do we need a separate temp table for sample taken date, separate from "temp_viral_load"?

configuration/pih/petl/jobs/hiv/viral_load/source.sql
```diff
@@ -0,0 +1,126 @@
...
+
+-- days since last viral load table
+CREATE TEMPORARY TABLE temp_days_since_vl
+(
+patient_id		INT,
+result_date		DATE
+);
+
+INSERT INTO temp_days_since_vl(patient_id, result_date)
+SELECT patient_id, MAX(vl_result_date) FROM temp_viral_load GROUP BY patient_id;
```
mogoodrich, line 61: Not sure of the business rules here, but shouldn't this be based on on the specimen collection date, not the result date?

Also, on a larger question, would this be the kind of thing better calculated in a downstream report by MEQ (ie in PowerBI, etc) rather than defined here?  This would make it easier for the MEQ to modify how it's calculated (and to be reminded how it's calculated, etc)

Interested in @dmdesimone 's thoughts as well.

configuration/pih/petl/jobs/hiv/viral_load/source.sql
```diff
@@ -0,0 +1,126 @@
...
+        viral_load,
+        vl_type,
+        DATEDIFF(NOW(), result_date) days_since_last_vl,
+        index_desc,
+        index_asc
+FROM temp_viral_load tvl
+-- speciment collection date 
+JOIN temp_vl_sample_taken_date tvld ON tvld.encounter_id = tvl.encounter_id 
+-- days since last viral load
+LEFT JOIN temp_days_since_vl tdl ON tvl.patient_id = tdl.patient_id AND tdl.result_date = tvl.vl_result_date
```
mogoodrich, line 121: So is the purposed of this join and table then to make sure the days since last viral load is calculated/displayed for only the most recent results for a patient? Or am I missing something?

configuration/pih/petl/jobs/hiv/viral_load/source.sql
```diff
@@ -0,0 +1,126 @@
...
+patient_id		INT,
+result_date		DATE
+);
+
+INSERT INTO temp_days_since_vl(patient_id, result_date)
+SELECT patient_id, MAX(vl_result_date) FROM temp_viral_load GROUP BY patient_id;
+
+-- index ascending
+DROP TEMPORARY TABLE IF EXISTS temp_index_asc;
+CREATE TEMPORARY TABLE temp_index_asc
```
mogoodrich, line 65: What does the index asc and index desc do?  Hard for me to follow.

mogoodrich (commented):
Thanks @jmbabazi ... I had a few questions which I've added.  Also interested in @dmdesimone 's thoughts on the questions.  Happy to discuss in person tomorrow.

configuration/pih/petl/jobs/hiv/viral_load/source.sql
mogoodrich: Ah, re: result date vs sample date, I see result date is the requirement from the Excel document.  Seems strange to  me, but maybe that's correct!

configuration/pih/petl/jobs/hiv/viral_load/source.sql
jmbabazi: sample taken date is the encounter date. easier way to bring it in is to have a separate table, other approaches throw sub query return more than one row error.

configuration/pih/petl/jobs/hiv/viral_load/source.sql
jmbabazi: defer to @dmdesimone about the calculations. I think this is what the excel file has

configuration/pih/petl/jobs/hiv/viral_load/source.sql
jmbabazi: Yes, see comments in excel file, introducing a new table is because I can't write to an open table. ideally a statement like 
update temp_viral_load set days_since_last_vl = datediff(now(), max(result_date)); would have been great but its not possible since  temp_viral_load  is open

configuration/pih/petl/jobs/hiv/viral_load/source.sql
jmbabazi: index asc or desc is counting a patient's encounters in ascending and descending order respectively

mogoodrich: Got it... not sure why we want this though.  If we want index by date, don't we need to sort by date?  Probably more a question about the underlying requirements, we can discuss with @dmdesimone  tomorrow...

configuration/pih/petl/jobs/hiv/viral_load/source.sql
jmbabazi: Done

configuration/pih/petl/jobs/hiv/viral_load/source.sql
jmbabazi: Done

configuration/pih/petl/jobs/hiv/viral_load/source.sql
jmbabazi: Done

configuration/pih/petl/jobs/hiv/viral_load/source.sql
dmdesimone: I agree!

configuration/pih/petl/jobs/hiv/viral_load/source.sql
dmdesimone: Per a separate discussion with Debbie, we agreed that yes, this should be Sample (encounter) date!

configuration/pih/petl/jobs/hiv/viral_load/source.sql
dmdesimone: yes, it should be ordered by sample (encounter date)

configuration/pih/petl/jobs/hiv/viral_load/target.sql
```diff
@@ -0,0 +1,14 @@
...
+        hiv_patient_id                      INT,
+        encounter_id                        INT,
+        vl_sample_taken_date                DATE,
+        vl_sample_taken_date_estimated      VARCHAR(255),
+        vl_result_date                      DATE,
+        vl_test_outcome                     VARCHAR(255),
+        vl_result_detectable                VARCHAR(255),
+        viral_load                          INT,
+        vl_type                             VARCHAR(255),
+        days_since_last_vl                  INT,
```
mseaton, line 11: Per above, call this "days_since_vl", and populate it for every row.

mseaton (commented):
Generally looks good @jmbabazi .  A few comments and suggestions.  @mogoodrich  FYI

configuration/pih/petl/jobs/hiv/viral_load/source.sql
```diff
@@ -0,0 +1,122 @@
...
+
+## Delete test patients
+DELETE FROM temp_hiv_construct_encounters WHERE
+patient_id IN (
+               SELECT
+                      a.person_id
+                      FROM person_attribute a
+                      INNER JOIN person_attribute_type t ON a.person_attribute_type_id = t.person_attribute_type_id
+                      AND a.value = 'true' AND t.name = 'Test Patient'
+               );
```
mogoodrich, line 44: Tangetially, I don't think I did this for the hiv_patients table, we should correct that there as well.

configuration/pih/petl/jobs/hiv/viral_load/source.sql
```diff
@@ -0,0 +1,122 @@
...
+        tvl.patient_id,
+        tvl.encounter_id,
+        DATE(tvl.vl_sample_taken_date),
+        vl_sample_taken_date_estimated,
+        vl_result_date,
+        vl_test_outcome,
+        vl_result_detectable,
+        viral_load,
+        vl_type,
+        DATEDIFF(NOW(), tvl.vl_sample_taken_date) days_since_vl,
```
mogoodrich, line 114: I prefer this much better to than the "days since last viral"!  Just want to make sure we confirm this change with the MEQ team though to avoid misinterpretation down the line.  (We should lobby hard for this way though IMO!)

mogoodrich (approved):
LGTM, with the one small comment that @mseaton made

configuration/pih/petl/jobs/hiv/viral_load/source.sql
jmbabazi: Done

PR #66, opened 2020-09-22
UHM-4812: Viral load warehouse table and ETL

mogoodrich:


configuration/pih/liquibase/sql/create_functions.sql
```diff
@@ -800,7 +800,7 @@ END
 #
 
 -- This function accepts encounter_id, mapping source, mapping code
--- It will find a single, best observation that matches this, and return the value_text
+-- It will find a single, best observation that matches this, and return the concept name
```
mogoodrich, line 803: Just changed this to clarify that is returns the concept_name of the value_coded, not the value_text (since value_text is something different)

configuration/pih/petl/jobs/hiv/viral_load/source.sql
```diff
@@ -116,8 +126,9 @@ SELECT
         vl_result_date,
         specimen_number,
         vl_test_outcome,
-        IF(vl_result_detectable IS NOT NULL, 1, NULL) vl_result_detectable,
+        IF(vl_test_outcome_coded = @detected_viral_load, 1, 0) vl_result_detectable,
```
mogoodrich, line 130: So I added vl_result_detectable back in, but it's now based on the coded vl_test_outcome.

mogoodrich (commented):
@dmdesimone  @jmbabazi  can you (re)review when you get a chance?  In particular, I'd like to rename the obs_value_coded_list_concept_id function that I created but wanted your thoughts on the best name.

PR #67, opened 2020-09-25
Add function for ED Triage rewrite

dmdesimone:
Add function for ED Triage rewrite (UHM-4876)

### DIFF
--- a/configuration/pih/liquibase/sql/create_functions.sql
+++ b/configuration/pih/liquibase/sql/create_functions.sql
@@ -1755,3 +1755,31 @@ BEGIN
 
 END
 #
+-- This function acceps patient id, a list of encounter types, visit id, form id and a begin date
+-- it will return the latest encounter for that patient, of one of those encounter types, in that visit of that form type since that date
+-- form id and begin date can be left null
+#
+DROP FUNCTION IF EXISTS latestEncForminVisit;
+#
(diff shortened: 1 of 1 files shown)


configuration/pih/liquibase/sql/create_functions.sql
```diff
@@ -1755,3 +1755,31 @@ BEGIN
 
 END
 #
+-- This function acceps patient id, a list of encounter types, visit id, form id and a begin date
+-- it will return the latest encounter for that patient, of one of those encounter types, in that visit of that form type since that date
+-- form id and begin date can be left null
+#
+DROP FUNCTION IF EXISTS latestEncForminVisit;
+#
+CREATE FUNCTION latestEncForminVisit(_patientId int(11), _encounterTypes varchar(255), _visitId int(11), _formId int(11), _beginDate datetime)
```
mogoodrich, line 1764: We should be a little careful of referencing encounter types by name instead of UUID, but guess the names are fairly stable.

I was suprised/had forgotten that we had used the same encounter type for Consult and ED Triage forms...

mogoodrich (approved):

configuration/pih/liquibase/sql/create_functions.sql
dmdesimone: Nice catch.  I meant to go back and replace the code that I used to call with a parameter I load in from the encounter type table.  I will make this change and commit!

PR #69, opened 2020-09-25
Changing from functions to inner join

jmbabazi:


configuration/pih/petl/jobs/hiv/viral_load/source.sql
```diff
@@ -18,11 +18,10 @@ CREATE TEMPORARY TABLE temp_hiv_construct_encounters
     vl_sample_taken_date_estimated  VARCHAR(11),
     vl_result_date                  DATE,
     specimen_number                 VARCHAR(255),
-    vl_test_outcome_coded           INT,
-    vl_test_outcome                 VARCHAR(255),
+    vl_coded_results                VARCHAR(255),
```
mogoodrich, line 21: Saw your comment on the PR... I didn't realize that "vl_test_outcome" was meant to be something else... I agree that this new name you use is better.

configuration/pih/petl/jobs/hiv/viral_load/source.sql
```diff
@@ -18,11 +18,10 @@ CREATE TEMPORARY TABLE temp_hiv_construct_encounters
...
     vl_result_detectable            INT,
     viral_load                      INT,
-    detected_lower_limit            INT,
+    ldl_value                       INT,
```
mogoodrich, line 24: Fine with me, this seems like a clearer name.

mogoodrich (approved):
LGTM, though I mainly just reviewed the names, etc... will let Dave take a look at the queries themselves.

PR #71, opened 2020-09-29
Changing varchar dtype to text

jmbabazi:


jmbabazi:
@dmdesimone , on workbench you can get the data type issue. So I thought it was easier to change the varchar(255) to text.

PR #72, opened 2020-10-02
Adding causing of death to the patient table

jmbabazi:


mogoodrich (commented):
LGTM!  I don't know the full requirements for reporting, but might make sense to add the "dead" column and potentially the "death_date" column as well.

Also, i don't think we allow entry of a "cause_of_death_noncoded" right now, but I could be wrong about that, and it doesn't seem harmful to have it.

mogoodrich (approved):

jmbabazi:
Thanks @mogoodrich. I've added the 2 new columns (dead and date of death). We don't collect cause_of_death_noncoded but I figure it wouldn't be harmful to have it,

mogoodrich:
Cool! Yeah, totally agree regarding "cause_of_death_noncoded"

PR #75, opened 2020-10-06
Adding translation for key population

jmbabazi:


configuration/messageproperties/messages_en.properties
```diff
@@ -1048,6 +1052,7 @@ pihcore.genitalUlcer=Genital ulcer
 pihcore.withoutCondom=Without condom
 pihcore.analSex=Anal sex
 pihcore.homoSexual=Male homosexuality activity
+pihcore.homosexual=Men who have sex with other men
```
lnball, line 1055: @jmbabazi I would recommend that you change the code for pihcore.homosexual.  It's confusing to have another one with different text but uppercase S "homoSexual".  Maybe change to menSexWithMen.  Thanks.

mogoodrich: Agree with @lnball

mogoodrich (changes requested):
Agree with @lnball 's changes...

lnball:
Sorry to be so picky, James.  I'll approve with the small change.  Then you'll also need to update the htmlform.

lnball (approved):

PR #78, opened 2020-10-07
Create target.sql

dmdesimone:


jmbabazi (changes requested):
One column is missing and the order does not match that of the select statement. @dmdesimone

PR #79, opened 2020-10-09
Key population ETL

jmbabazi:


mogoodrich (approved):
@jmbabazi  I looked this over and it looks good to me, though I'd defer to @dmdesimone  since he's better with SQL than me as to how performant this would be.

PR #88, opened 2020-11-17
UHM-4979 OVC Intake form

brandones:
Screenshot from 2020-11-16 17-48-47
Screenshot from 2020-11-16 17-49-49
Screenshot from 2020-11-16 17-50-07


brandones:
The form entry is working, but what's weird is that after saving the form, nothing shows up in view mode. An encounter is created and the enrollInProgram happens, but nothing is shown in the visit note.

Screenshot from 2020-11-16 18-29-01

configuration/pih/scripts/visit/encounterTypeConfig.js
```diff
@@ -1107,6 +1107,31 @@ angular.module("encounterTypeConfig", [])
             ]
           }
         };
+        
+        // outpatientConsult
+        encounterTypes["92fd09b4-5335-4f7e-9f63-b2a663fd09a6"] = {
```
mogoodrich, line 1112: This is tangential, I assume?

configuration/pih/htmlforms/ovcIntake.xml
```diff
@@ -0,0 +1,265 @@
+
+<htmlform
+    formUuid="a2f4e400-df59-4456-935f-cd0697ff28a4"
+    formName="OVC Intake"
+    formEncounterType="651d4359-4463-4e52-8fde-e62876f90792"
+    formVersion="1.0">
+
+    <div style="display: none">
+        <encounterDate default="today" />
+    </div>
```
mogoodrich, line 10: Are we allowing the encounter date to be changed at all?

mogoodrich (changes requested):
@brandones  I suspect that the encounter is not getting associated with the visit because there's no encounter location tag on the form.

I'd take a look at the existing code setting the encounter date, provider and location in the other forms.  For example, see oncologyConsult.xml:189-303.   There are three separate blocks of code based on the privileges of the user that came from the original requirements Mirebalais... these may have been too complicated and not necessary for this for, but at minimum I think we'd want the basic abilities to edit the date and location (oncologyConsult.xml:198-223).

configuration/pih/htmlforms/ovcIntake.xml
brandones: Replaced with the block you recommended

configuration/pih/scripts/visit/encounterTypeConfig.js
brandones: Whoops! That should have gone in a while ago, with the Peru stuff. Opened https://github.com/PIH/openmrs-config-pihemr/pull/89

brandones:
@mogoodrich Thanks for the advice. I added the block you recommended. Getting closer, but still not quite right in view mode:

Screenshot from 2020-11-17 09-58-18

Any idea why the form content wouldn't show up? There's nothing in the logs or the error console.

brandones:
Ping about the question above, @mogoodrich ?

mogoodrich:
This should be driven by the encounterTypeConfig, and since I don't think we have a special config for this encounter type, I would think would use the default, as defined here.

https://github.com/PIH/openmrs-config-pihemr/blob/master/configuration/pih/scripts/visit/encounterTypeConfig.js#L388

"defaultstate" means whether it shows opened or closed by default, and "shortTemplate" is the template to use  when it's closed, while "longTemplate" is what it's used when it's open.  And "templateModelUrl" is the endpoint it needs to call to get the data to back the template.

So I'd say it either isn't picking up the DEFAULT properly, or the default isn't working with this form.

Most of the code that drives all this is found here, so you could likely put in some breakpoints to see what is going on:
https://github.com/PIH/openmrs-module-pihcore/blob/master/omod/src/main/webapp/resources/scripts/visit/visit.js

Maybe add a breakpoint here and see what template is being set?
https://github.com/PIH/openmrs-module-pihcore/blob/master/omod/src/main/webapp/resources/scripts/visit/visit.js#L180

If you get really blocked feel free to merge and assign to me to debug...

brandones:
@mogoodrich Got it working. `encounterTypeConfig` wasn't using the DEFAULT configuration. Fixed in https://github.com/PIH/openmrs-config-pihemr/pull/88/commits/ac5494f8d4b56b04de3236ab805e3821520cef57 .

Screenshot from 2020-11-18 12-28-00

configuration/pih/scripts/visit/encounterTypeConfig.js
```diff
@@ -1165,7 +1165,7 @@ angular.module("encounterTypeConfig", [])
             var encounterType = encounterTypes[uuid];
 
             if (encounterType == null) {
-              return null;
```
mogoodrich, line 1168: Oh, cool, got, thanks!  I guess we must never have relied on the default before now?

mogoodrich (approved):
LGTM, thanks @brandones !

configuration/pih/scripts/visit/encounterTypeConfig.js
brandones: Seems like it, yeah

PR #89, opened 2020-11-17
PER-371: Consult note form - add outpatientConsult config for Peru

brandones:
https://pihemr.atlassian.net/browse/PER-371


mogoodrich (approved):
Cool, LGTM, let's just make sure we test on HUM-CI that the main consult note is still working properly after this change (the smoke tests *should* catch it, but...)

aastupilloses:
Here, tbh, I'm not 100% sure about the use of this, but as far as I can see, it's OK.

brandones:
@aastupilloses This is what specifies that, for Peru, the outpatient consult form will have those four sections.

Thanks for taking a look. I'll merge.

PR #90, opened 2020-11-17
UHM-4985 OVC follow up form

brandones:
https://pihemr.atlassian.net/browse/UHM-4985

See also https://github.com/PIH/openmrs-config-pihemr/pull/88

### DIFF
--- a/configuration/pih/htmlforms/ovcFollowup.xml
+++ b/configuration/pih/htmlforms/ovcFollowup.xml
@@ -1,0 +1,264 @@
+
+<htmlform
+    formUuid="5f486bc6-b017-4932-8037-6d667f8c3162"
+    formName="OVC Follow-up"
+    formEncounterType="f8d426fd-132a-4032-93da-1213c30e2b74"
+    formVersion="1.0">
+
+    <style type="text/css">
+
+        #who-when-where {
+            margin-bottom: 6px;
+            border-bottom: 1px solid #ccc;
+        }
+
+        #who-when-where p {
+            display: inline-block;
+            padding-right: 20px;
+        }
+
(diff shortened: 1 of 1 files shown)


configuration/pih/htmlforms/ovcFollowup.xml
```diff
@@ -0,0 +1,264 @@
...
+        </p>
+        <p><uimessage code="last_updated_on"/>:
+            <lookup complexExpression="$form.dateChanged"/>
+        </p>
+        <p><uimessage code="printed_on"/>:
+            <lookup complexExpression="$formGeneratedDatetime"/>
+        </p>
+    </div>
+
+    <includeIf velocityTest="$user.hasPrivilege('Task: emr.retroConsultNote')">
```
mogoodrich, line 176: It's probably fine to just have this one block for the date/location/provider instead of the three we've been using elsewhere, but then there should be no includeIf around this block... ie, we need to make sure the "who-when-where" *always* appears on the form, but it's probably fine to have this version, that allows editing all the three values, displayed for all users, instead of all the conditional stuff.  Could check with Dave or Ellen.  Also, same goes for the Intake form.

mogoodrich (changes requested):
Defer to Ellen on the form content, but one comment/change on the provider/location/date section.

configuration/pih/htmlforms/ovcFollowup.xml
lnball: This is much cleaner.  That section kept growing and was duplicated too many times.  Thanks @mogoodrich .  I'll approve.

lnball (approved):

PR #91, opened 2020-11-19
fix how unknown patient function us used

dmdesimone:
fix how unknown patient function us used - UHM-5032


jmbabazi:
Thanks @dmdesimone , this looks good. I am merging this PR

PR #94, opened 2020-11-23
Fix location tag for OVC forms

brandones:


mogoodrich (changes requested):
Why this change?  I think Consult Note location was (more) correct... 

A simplified view of the Haiti model might be something like this:

Mirebalais Hospital (Visit Location)
- Ward 1 (Consult Location)
- Ward 2 (Consult Location)

Cange (Visit Location, Consult Location)

Hince (Visit Location, Consult Location)

*If* things are working correctly, with the recent new global property "Restrict Encounter Location To Current Visit Location" set to true, it would work like this:

The initial pass would restrict the list of encounter locations to those tagged as Consult Locations:

Ward 1
Ward 2
Cange
Hinche

Then, if there's a visit in the context (which there should be in all our use cases) then it should further restrict the list to the current visit location OR it's children/grandchildren/etc).

So for a visit at Mirebalais we'd end up with the  following options:

Ward 1
Ward 2

And for a visit at Cange we'd end up with:
Cange

And for a visit at Hinche:

Hinche

The queston for me would be whether we want to use "Consult Location" or if there's a "HIV Consult Location" we want to use, or if we'd need to create a new "OVC Location"...

Happy to explain further...

brandones:
Ok. Thanks for the explanation, Mark. I'll add the Consult Location tag to all the locations in `pih-config-hiv.json`.

mogoodrich:
@brandones  it's likely fine to just use the "HIV Consult Location" which looks like has already been added in pih-config-haiti-hiv.json... defer to @lnball  and @dmdesimone

lnball:
This is a question for ZL (Ricard)?  Not sure if this would be completed by the HIV clinic or pediatrics...

PR #99, opened 2020-12-08
Make HIV Intake form a non-sectioned form for Peru

brandones:


mogoodrich (approved):
I forget 100% the conditional format we set up for default and an override, but I'll assume you remembers... :)  LGTM...

aastupilloses:
I'm not 100% sure about this, but is not "...encounterTypes.DEFAULT" invalid syntax for JS?

brandones:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment since ES6.

PR #104, opened 2020-12-22
Peru translation

MiguelAHPpih:
Made a new branch for translations. Added a new string for lastName2 and translations for the COVID 19 for that are currently WIP.


brandones:
@MiguelAHPpih You pulled the translation file with `tx pull`?

@mogoodrich This is how we manage translations, right? Manually pulling from Transifex and committing the translation files? Or does Transifex do the committing?

MiguelAHPpih:
> @MiguelAHPpih You pulled the translation file with `tx pull`?
> 
> @mogoodrich This is how we manage translations, right? Manually pulling from Transifex and committing the translation files? Or does Transifex do the committing?

@brandones There is a special commit command for translations? I tought i had to make the commit here and then it would appear on that translation page you showed us a month ago.

mogoodrich:
@brandones  you are correct, we manually pull in the translations from Transifex and commit.  It's possible to set up Transifex to automatically commit, I believe, but we haven't done that.

FWIW, if we are sure no other changes have been made in Transifex, we can force-push the messages_es.properties file up to Transifex, but would need to be careful to not blow away changes Brandon or others may have made.

(Side note, also looks like there's a conflict between the encounter type config in this PR and the other outstanding open one).

brandones:
Sounds good. @MiguelAHPpih please resolve the conflicts and we can merge.

aastupilloses:
Hi, guys! I have resolved the conflicts. Can you check if changes are OK? Thanks!

brandones:
Hi @aastupilloses and @MiguelAHPpih . I just re-read Miguel's comment (and we've talked some on Teams). As I expressed to him, translation files are managed [as described in the README](https://github.com/PIH/openmrs-config-pihemr/#message-properties).

> Note that the only messages properties file that should be edited directly in the code is the source file, messages_en.properties. For translations we use a third-party translation tool, Transifex (www.transifex).
>
> To provide translations, you need to add new codes to the messages_en.properties file, push them up to Transifex, do the translations, and then pull down the translations.

Does one of you have time to put these translations in Transifex today?

The problem with putting the translations directly into the messages file here is that the next time someone updates the messages files from Transifex, all of these translations will be wiped out, since Transifex doesn't know about them.

brandones:
I'm merging this. We'll push the messages_en file to Transifex, add translations there, then pull the messages_es file and commit to git.

PR #114, opened 2021-01-21
UHM-5092, make Specimen number top level obs

cioan:


mogoodrich (commented):
Generally looks good to me, but just remembered... @lnball  are we using this form anywhere in production and do we have to worry about any migration (ugh!)

configuration/pih/htmlforms/labResults.xml
```diff
@@ -159,7 +159,18 @@
...
-
+            <br/>
+            <p>
+                <label for="sample-id">
+                    <uimessage code="pihcore.lab.sample_id"/>
+                </label>
+                <span class="small">
+                    <field>
+                        <obs  id="sample-id" conceptId="CIEL:162086"
+                              placeholder="Enter Sample ID"/>
```
lnball, line 170: @mogoodrich I didn't follow the conversation today post-standup about specimen id.  It appears that the concept is the same from this file.  What changed for the data/concept?  (Team me if that's easier)

cioan: Just talked to @lnball , and we are good. No one is using this form yet. Thanks!

PR #115, opened 2021-01-22
enable covid reports in Peru

mogoodrich:


configuration/reports/reportdescriptors/dataexports/covid19dailycensus.yml
```diff
@@ -17,5 +17,6 @@ config:
   order: 140
   components:
     - "covid19"
+    - "covid19IntakeForm"
     - "allDataExports"
```
mogoodrich, line 21: I think this is the issue... in the yml for each report, we list the components which, if enabled, trigger that report to be included.

So, we wanted the covid report to be enabled if either the "covid19" component was enabled, or the "allDataExports" was enabled... but it looks like in Peru we have a special component called "covid19IntakeForm" that we set up separate from "covid19" (see https://github.com/PIH/openmrs-config-ses/blob/master/configuration/pih/pih-config-peru.json#L11) , so I think we need to add that... (haven't tested this though)...

brandones (approved):
Nice, makes sense. Glad to know how this works now.

PR #116, opened 2021-01-25
UHM-5119, show encounter type on visit list

cioan:


configuration/pih/scripts/visit/encounterTypeConfig.js
```diff
@@ -413,7 +413,7 @@ angular.module("encounterTypeConfig", [])
                 shortTemplate: "templates/encounters/defaultEncounterShort.page",
                 longTemplate: "templates/encounters/defaultHtmlFormEncounterLong.page",
                 templateModelUrl: "/module/htmlformentry/encounter.json?encounter={{encounter.uuid}}",
-                showOnVisitList: false
+                showOnVisitList: true
```
mogoodrich, line 416: Since we are changing this to true, for clarity, let's remove all the places where we specifically set this to true on the other encounter types.

Also, I think Dave asked for "checkIn" and "patientregristation" not to be shown, so those encounters should actually set "showOnVisitList" false...

mogoodrich (changes requested):

configuration/pih/scripts/visit/encounterTypeConfig.js
cioan: Thanks @mogoodrich ! Sorry I merged the PR before I saw your comment. I will update the code. Thanks!

mogoodrich: No worries, thanks!

PR #126, opened 2021-02-12
Concept_Name function should use preferred name first

dmdesimone:
Concept_Name function should use preferred name first

### DIFF
--- a/configuration/pih/liquibase/sql/create_functions.sql
+++ b/configuration/pih/liquibase/sql/create_functions.sql
@@ -72,7 +72,9 @@ BEGIN
 	WHERE voided = 0
 	  AND concept_id = _conceptID
 	  AND locale = _locale
-	  AND concept_name_type = 'FULLY_SPECIFIED';
+ 	order by locale_preferred desc, ISNULL(concept_name_type) asc, 
+	field(concept_name_type,'FULLY_SPECIFIED','SHORT')
+	limit 1;
 
     RETURN conceptName;
 END
--- a/configuration/pih/liquibase/sql/create_functions.sql
+++ b/configuration/pih/liquibase/sql/create_functions.sql
@@ -72,8 +72,9 @@ BEGIN
 	WHERE voided = 0
 	  AND concept_id = _conceptID
 	  AND locale = _locale
- 	order by locale_preferred desc, ISNULL(concept_name_type) asc, 
-	field(concept_name_type,'FULLY_SPECIFIED','SHORT')
+	order by if(_locale = locale, 0, 1), if(locale = 'en', 0, 1),
+	  locale_preferred desc, ISNULL(concept_name_type) asc, 
+	  field(concept_name_type,'FULLY_SPECIFIED','SHORT')
 	limit 1;
 
     RETURN conceptName;


configuration/pih/liquibase/sql/create_functions.sql
```diff
@@ -72,7 +72,9 @@ BEGIN
 	WHERE voided = 0
 	  AND concept_id = _conceptID
 	  AND locale = _locale
-	  AND concept_name_type = 'FULLY_SPECIFIED';
+ 	order by locale_preferred desc, ISNULL(concept_name_type) asc, 
+	field(concept_name_type,'FULLY_SPECIFIED','SHORT')
+	limit 1;
```
mseaton, line 78: @dmdesimone - I think this approach looks fine.  That said, the previous function was written under the assumption that there was always a Fully Specific name in each locale, and this could be relied upon to be there.  If that's not true, we likely need to do more to ensure there is _always_ some sort of name returned for a concept, and that likely requires looking beyond the given locale.  

What I'd say is that we should move locale out of a constraint and into another condition in the order by clause.  So, the first order by criteria would be something like if(_locale = locale, 0, 1).

This way, we basically get the first concept name for a concept, using our preferred order, but don't return null if a concept happens to be missing a name in the locale of choice.

dmdesimone: This is a good idea.  Extending this, I'd like to put if(locale='en',0,1) as the second sort criteria so it prioritizes English if there is no name in the passed-in locale.

configuration/pih/liquibase/sql/create_functions.sql
```diff
@@ -72,8 +72,9 @@ BEGIN
 	WHERE voided = 0
 	  AND concept_id = _conceptID
 	  AND locale = _locale
```
mseaton, line 74: @dmdesimone - it looks like you merged this in, but you never removed the locale constraint here.  You'll need to move the "AND locale = _locale" constraint for this to work as intended.

configuration/pih/liquibase/sql/create_functions.sql
```diff
@@ -72,8 +72,9 @@ BEGIN
...
- 	order by locale_preferred desc, ISNULL(concept_name_type) asc, 
-	field(concept_name_type,'FULLY_SPECIFIED','SHORT')
+	order by if(_locale = locale, 0, 1), if(locale = 'en', 0, 1),
+	  locale_preferred desc, ISNULL(concept_name_type) asc,
```
mseaton, line 76: Should I assume that the ISNULL returns 0 if not null, 1 if null, so names with non-null types are ordered before names with null types?  Is that the intent?

configuration/pih/liquibase/sql/create_functions.sql
```diff
@@ -72,8 +72,9 @@ BEGIN
 	WHERE voided = 0
 	  AND concept_id = _conceptID
 	  AND locale = _locale
- 	order by locale_preferred desc, ISNULL(concept_name_type) asc, 
-	field(concept_name_type,'FULLY_SPECIFIED','SHORT')
+	order by if(_locale = locale, 0, 1), if(locale = 'en', 0, 1),
+	  locale_preferred desc, ISNULL(concept_name_type) asc, 
+	  field(concept_name_type,'FULLY_SPECIFIED','SHORT')
```
mseaton, line 77: This is a cool function - I didn't know about it.  So this works with the ISNULL above to return FULLY SPECIFIED first, SHORT second, and then anything else after (assuming ordering is already the same).

PR #130, opened 2021-02-23
Changing date format to allow exports to ETl

jmbabazi:


dmdesimone (approved):
Looks good - will merge!

PR #131, opened 2021-02-26
Add TB Intake encounter type config

brandones:


mogoodrich (approved):
didn't check the uuid :)  but looks right to me!

PR #149, opened 2021-03-17
Add message codes for new occupation options

brandones:
https://pihemr.atlassian.net/browse/UHM-5234

Used in https://github.com/PIH/openmrs-module-mirebalais/pull/137


brandones:
Merging so that I can do the Transifex work.

PR #153, opened 2021-03-31
change view mode of OVC and TB encounter types to render using the HT…

mogoodrich:
…ML form


mogoodrich (commented):
This allows these forms to be rendered and laid out using the HTML Form template... (it's possible we want to just make this the default... I don't currently remember the use case for the current default, which also relies on HTML Form Entry, but appears to just do a "obs dump" instead of laying the data using the form.

brandones:
Sure. Heads up @MiguelAHPpih @aastupilloses . If Peru wants the TB Intake to look different in View mode, we can override for Peru.

brandones (approved):

mogoodrich:
Thanks @brandones !

PR #155, opened 2021-04-05
LIB-140 Updated liquibase for Liberia NCD forms

lnball:
For Liberia only so it will ignore the NCD Intake and Followup forms.

### DIFF
--- a/configuration/pih/liquibase/liquibase.xml
+++ b/configuration/pih/liquibase/liquibase.xml
@@ -125,8 +125,7 @@
 
             -- Ignore for Liberia NCD encounters when value_coded is NULL
             delete from temp_value_drug
-             where value_coded is NULL
-               and 'Liberia' IN (select name from address_hierarchy_entry where parent_id is NULL)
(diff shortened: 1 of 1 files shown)


configuration/pih/liquibase/liquibase.xml
```diff
@@ -120,6 +123,12 @@
                set temp_value_drug.value_coded = drug.concept_id
              where temp_value_drug.value_drug = drug.drug_id;
 
+            -- Ignore for Liberia NCD encounters when value_coded is NULL
+            delete from temp_value_drug
+             where value_coded is NULL
```
mogoodrich, line 128: Don't we want to skip the NCD Intake and NCD FU for Liberia *entirely*.   On this form only value-coded are collected, right?  what if it possible that the value_drug does have that same primary key, and maps back to the another concept, although it was correct in the first place?

lnball: You're right.  I thought of that, but didn't update the script.  Will fix.

PR #162, opened 2021-04-13
UHM-4354 Move Location, Location Tag, and Lcation Attribute Types installation out of PIH core and into config repos

brandones:
https://pihemr.atlassian.net/browse/UHM-4354

### DIFF
--- a/configuration/locations/core-locations.csv
+++ b/configuration/locations/core-locations.csv
@@ -1,2 +1,3 @@
 UUID,Void/Retire,Name,Description
 AF939782-898B-409A-99A4-D1653484EDBD,,Patient Home,The patient is visited at home
+8d6c993e-c2cc-11de-8d13-0010c6dffd0f,,Unkown Location,Unknown Location
--- a/configuration/locations/core-locations.csv
+++ b/configuration/locations/core-locations.csv
@@ -1,0 +1,2 @@
+UUID,Void/Retire,Name,Description
+AF939782-898B-409A-99A4-D1653484EDBD,,Patient Home,The patient is visited at home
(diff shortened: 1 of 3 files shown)


configuration/locations/core-locations.csv
```diff
@@ -0,0 +1,2 @@
+UUID,Void/Retire,Name,Description
+AF939782-898B-409A-99A4-D1653484EDBD,,Patient Home,The patient is visited at home
```
mogoodrich, line 2: What about "Unknown Location"? Are we assuming that is part of the core OpenMRS install?

mseaton: I would assume so, but doesn't hurt to include it for completeness.  I'd vote to include it here.

mseaton (approved):
I did not review the details of the csvs - hopefully you got all of those right and got them by exporting from an existing DB, as that leaves less room for transcription errors.

configuration/locations/core-locations.csv
brandones: Initially I had it included and it caused an NPE in concept validation, something about its location attributes being null. Chose not to worry about it.

brandones:
I've been using vim, since all this stuff is in code already. Just a few `g/` and `%s/` commands to clean the MDD version into a valid CSV.

configuration/locations/core-locations.csv
mogoodrich: Strange... what was the error @brandones

brandones: Huh, it worked this time when I re-ran it. Added that line back. We'll see if it causes trouble.

PR #164, opened 2021-04-15
UHM-5428 - Add tb screening result to status form, to show only if th…

mseaton:
…e result is not null


mogoodrich (approved):
LGTM, merging!

mogoodrich:
I guess we are saying since this is a very specific use case, we will just always use the French translation?

PR #165, opened 2021-04-16
med-order variable for enconterType

BranfordTGbieor:
fyi @lnball


lnball (approved):
LGTM (looks good to me ;)

PR #166, opened 2021-04-21
UHM-4425, install app privileges via iniz config

cioan:


mogoodrich (approved):
Didn't double-check them all, but looks right.  I'd just fire up a system from scratch to test if you haven't already.

One nice thing is that since you are just moving the Privileges to start, if there are any typos/etc, I think the Roles install would fail hard to alert you?

PR #169, opened 2021-04-24
Fix Attribute Type CSV. Would not load on a fresh server without 'min…

brandones:
… occurs' and 'max occurs'


mogoodrich (approved):

brandones:
Should fix https://pihemr.atlassian.net/browse/UHM-5452

PR #170, opened 2021-04-26
Fix cange

brandones:
Preparing for UHM-4354, https://pihemr.atlassian.net/browse/UHM-4354

When run on a new server, Iniz will add Cange with the new UUID and this changeset will not run.

When run on an already-existing server, Iniz will first add all the locations but fail to add the fixed Cange location, complaining that a location named Cange already exists. Then this liquibase script will run and update the existing Cange location.

### DIFF
--- a/configuration/pih/liquibase/liquibase.xml
+++ b/configuration/pih/liquibase/liquibase.xml
@@ -154,4 +154,28 @@
         </sql>
     </changeSet>
 
+    <changeSet id="20210426-fix-cange-uuid" author="bistenes">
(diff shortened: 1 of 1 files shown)


mogoodrich:
@brandones so when iniz can't install a piece of metadata, it logs a warning and moves on?  (Seems reasonable vs our approach of stopping everything dead, but good to know).

mogoodrich (approved):

brandones:
Correct, yes.

PR #172, opened 2021-04-29
Create delivery.sql

dmdesimone:
Create delivery.sql


mseaton (approved):
From a cursory code review perspective looks good to me @dmdesimone .  Testing should highlight any issues with the actual logic in the query.

PR #173, opened 2021-04-29
Create delivery.yml

dmdesimone:
Create delivery.yml


mseaton (approved):
@dmdesimone this looks fine, but you should avoid creating multiple PRs and multiple branches for the same commit.  You can continue to push commits to the same branch and open a single PR against it, and if a PR is already open it will update the PR.

dmdesimone:
Thanks Mike - next set of PRs I will try that.  (I use the web UI for github so it might not be so obvious)

PR #175, opened 2021-04-29
add delivery message properties

dmdesimone:
add delivery export message properties


mseaton (approved):
Fine I guess, but do we modify these like this or do we use transifex?  Why are you creating PRs for every file you are modifying?

PR #184, opened 2021-05-10
PIH Concept Sources MDS to 5 - adds 'SES Lab'

brandones:
Peru wants to attach codes from 6Labs (their LIMS) to all their lab concepts. I'd like to define a new concept source that they can attach these codes to.


mogoodrich (approved):
Seems fine to me @brandones ... and yes, this would mean we'd have to do "deferred" loading of concepts, but I think this makes sense and shouldn't be hard to do...

mseaton:
If it were me, this would be my opportunity to bang support for a "conceptsource" domain into Iniz and push that through.  Why not take that route?

brandones:
@mseaton Yeah, that's a good idea. I'm gonna merge this for now since I don't want Peru blocked on a Mekom review. I'll file a ticket for a concept source domain and see if it comes together while I'm waiting for builds.

PR #185, opened 2021-05-17
UHM-5517: Unable to Save Some Forms when Visit Location does not match Session Location

mogoodrich:
So, now that we've got situations where there are multiple visit locations on a single server, we have an issue if the user's session location does not match the visit location.

As a workaround for this, we will *always* allow users to change the encounter location.

This is basically a sample of the change made to COVID Admission form for review, which I'll then have to cut-and-paste to all other forms.

It also includes a change to the Vitals form, which as a different layout.

### DIFF
--- a/configuration/pih/htmlforms/covid19Intake.xml
+++ b/configuration/pih/htmlforms/covid19Intake.xml
@@ -263,7 +263,7 @@
         </includeIf>
 
         <!-- all users that don't have either retro privilege, or those with retro-this-provider-only but with an active visit, can only edit location -->
-        <includeIf velocityTest="(!$user.hasPrivilege('Task: emr.retroConsultNoteThisProviderOnly') and !$user.hasPrivilege('Task: emr.retroConsultNote')) or ($user.hasPrivilege('Task: emr.retroConsultNoteThisProviderOnly') and $visit.open)">
+        <includeIf velocityTest="(!$user.hasPrivilege('Task: emr.retroConsultNoteThisProviderOnly') and !$user.hasPrivilege('Task: emr.retroConsultNote')) or ($user.hasPrivilege('Task: emr.retroConsultNoteThisProviderOnly') and !$user.hasPrivilege('Task: emr.retroConsultNote') and $visit.open)">
             <div style="display:none">
                 <encounterProviderAndRole default="currentUser" encounterRole="4f10ad1a-ec49-48df-98c7-1391c6ac7f05"
                                           required="true"/>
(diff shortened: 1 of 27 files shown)


configuration/pih/htmlforms/covid19Intake.xml
```diff
@@ -262,12 +262,11 @@
 
         </includeIf>
 
-        <!-- all users that don't have retroConsultNote privilege cannot edit provider, location or date when active visit -->
-        <includeIf velocityTest="(!$user.hasPrivilege('Task: emr.retroConsultNote')) and ($visit.open)">
```
mogoodrich, line 266: So while I was reviewing the three possible scenarios, I discovered a bug... if a user did not have either of the retro consult note privileges and the visit would not open, none of three encounter details elements would be used, and there's be a stack trace when saving the form.

This was because (I assume?) in our original design you never should have been able to reach that case... ie, you shouldn't have been able to add a retro note if you didn't have either of the retro privileges.  But we'd gotten away from this.

configuration/pih/htmlforms/vitals.xml
```diff
@@ -461,6 +461,35 @@
...
+                                <encounterLocation default="SessionAttribute:emrContext.sessionLocationId"
+                                                   tags="Vitals Location"/>
+                            </field>
+                        </p>
+                    </span>
+                    <div  style="display:none">
+                        <encounterDate id="encounterDate" default="now"/>
+                    </div>
+                </fieldset>
+            </includeIf>
```
mogoodrich, line 492: This will add another question to the "Vitals" note for all users, asking for the location.  Hopefully we are okay with this.

mogoodrich (commented):
I need to do some more testing of this tomorrow, but wanted to get it out for review... I'll have to copy this across to all forms in the system, more or less.

mogoodrich:
I'd still love it if we could simplify this... :)

configuration/pih/htmlforms/vitals.xml
lnball: @mogoodrich Will it default the location on the vitals note?  Or you need to set it?  Otherwise it seems annoying for the nurse auxilliary who just collect vitals at HUM...

mogoodrich: Yes, good point @lnball ... it will default to the proper location... the exception would be if the session location doesn't match the visit location (which would never happen for Mirebalais, only for systems that have multiple visit locations like Haiti HIV).

brandones:
Whatever you do here is fine by me. Mexico doesn't have any permissions issues—anyone that can fill a consult note can edit one.

mseaton (approved):
LGTM, thanks @mogoodrich

PR #186, opened 2021-05-18
UHM-5528: Move encounter types into config

mseaton:


constants.yml
```diff
@@ -27,19 +27,122 @@ encounterRole:
 
 # for easy lookup, keys should be the name of the encounter type in the DB
```
mogoodrich, line 28: Why the change from camel case to all uppercase with underscore?  To make it more clearly map to the name in the DB?

mogoodrich (approved):
LGTM!  Curious about the change to all-caps and underscores...

constants.yml
mseaton: These were basically the names of the constant variables in the EncounterTypes.java file, and it was easier to make all of the changes in the code by leaving them alone.  Also, it seemed to align with previous comment that you made about having the plugin add underscores where there was camel case to make the constants more readable - I ended up just changing the constant names to achieve that here.  Thoughts?

PR #187, opened 2021-05-18
Add default frontend config to config-pihemr

brandones:
This should make it so that the frontend doesn't fail to load due to lack of a config file, even for servers where the country config repo doesn't have a frontend config file. And the presence of a country config file will override this one.


brandones:
(no code to review here, @mseaton , just want to double-check my logic about config inheritance)

mseaton:
That’s correct @brandones - that’s how the inherits should work.

mogoodrich:
Sounds correct to me @brandones

PR #198, opened 2021-05-28
add concept_in_set function

dmdesimone:
add concept_in_set function


mseaton (approved):
This looks generally good to me @dmdesimone .  It'd be interesting to see if you can accomplish this with recursion, where this function can call itself if there is a parent set until both no match is found and no parent is found to check.  That would give you full depth searching.  But this looks fine.

PR #202, opened 2021-06-02
Remove unused programs

jmbabazi:


mseaton (approved):
This looks good to me @jmbabazi , assuming these programs have no workflows or states to look at.  I think we can put this in the main pihemr config, as the preconditions will ensure it doesn't run where these programs are in use, and the delete statement will not have an adverse impact if it runs where these programs are not present.

configuration/pih/liquibase/liquibase.xml
```diff
@@ -154,4 +154,17 @@
...
     </changeSet>
 
+    <changeSet id="20210602-delete-unused-ces-programs" author="jmbabazi">
+        <preConditions onFail="MARK_RAN">
+        <sqlCheck expectedResult="0">
+            select count(*) from patient_program where program_id in
+            (select program_id from program where uuid in ("9fe0c32c-981f-4283-b4f4-168285a52c7d", "40657f91-5991-44fe-a16f-220dbeb9dcbd", "1eb13158-f39c-444a-b77c-78f1d78123be"));
+        </sqlCheck>
+        </preConditions>
+        <comment>If the count above is 0, then remove these programs (Hypertension program, Diabetes program, Epilepsy program)</comment>
```
mogoodrich, line 164: We only want to remove these programs if the count of patient programs = 0, not above 0?

(It looks like the code is doing the right thing, but the comment is wrong?)

jmbabazi: Thanks @mseaton , @mogoodrich . I removed the above from the comments

configuration/pih/liquibase/liquibase.xml
```diff
@@ -154,4 +154,37 @@
...
+        <preConditions onFail="MARK_RAN">
+            <sqlCheck expectedResult="0">
+            select count(*) from patient_program where program_id in
+            (select program_id from program where uuid in ("9fe0c32c-981f-4283-b4f4-168285a52c7d",
+                "40657f91-5991-44fe-a16f-220dbeb9dcbd", "1eb13158-f39c-444a-b77c-78f1d78123be") and retired = 0);
+            </sqlCheck>
+        </preConditions>
+        <comment>If the count is 0, then remove these programs (Hypertension program, Diabetes program, Epilepsy program)</comment>
+        <sql>
+            -- void patient states related to these programs
```
mseaton, line 167: This does not make sense to me @jmbabazi .  We can't void states, and there won't be states anyway.  If the precondition passes (there are no patient programs), then by definition there can't be any patient states associated with these programs, right?  And if there were, voiding them and then deleting the states in the next statement doesn't work.

mseaton (approved):
LGTM

jmbabazi:
Thanks @mseaton

PR #203, opened 2021-06-02
UHM-5548:  Migrate concept sources to Initializer

mseaton:


mogoodrich (approved):
Didn't compare the concept source directly, but we can do that during testing, otherwise LGTM...

PR #204, opened 2021-06-04
Add Diagnosis functions

dmdesimone:
Add 5 diagnosis functions

### DIFF
--- a/configuration/pih/liquibase/sql/create_functions.sql
+++ b/configuration/pih/liquibase/sql/create_functions.sql
@@ -1945,3 +1945,131 @@ limit 1;
 
 END
 #
+-- The following diagnosis functions can work together to provide all of the details of diagnoses entered in encounters
+-- they each accept encounter_id and offset and will return the details of the nth (based on offset) encounter,
+-- ordered by: 
+--   primary diagnoses first, then other diagnoses with an order, then diagnoses without an order. obs_id is included as a final sort factor
+#
+-- The following function accepts encounter_id and offset
+-- It will return the obs_group_id of the construct of the nth diagnosis of that encounter (n being the offset)
+#
+DROP FUNCTION IF EXISTS diagnosis_obs_group_id;
+#
+CREATE FUNCTION diagnosis_obs_group_id(_encounter_id int(11), _offset int(11)) RETURNS int(11)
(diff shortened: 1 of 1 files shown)


configuration/pih/liquibase/sql/create_functions.sql
```diff
@@ -1945,3 +1945,131 @@ limit 1;
...
+BEGIN
+
+DECLARE ret int(11);
+
+select ogr.obs_id into ret from obs ogr 
+ left outer join obs oo on oo.obs_group_id = ogr.obs_id and oo.concept_id = concept_from_mapping('PIH','7537') and oo.voided = 0
+where ogr.encounter_id = _encounter_id
+and ogr.concept_id = concept_from_mapping('PIH','7539')
+and ogr.voided = 0
+order by ISNULL(oo.value_coded) ASC ,LOCATE('7534',retrieveConceptMapping(oo.value_coded,'PIH')) DESC, ogr.obs_id ASC
```
mseaton, line 1969: So I guess this works if there are exactly 2 choices for order - primary and secondary - but if additional options were available we'd have to tweak it, right?

configuration/pih/liquibase/sql/create_functions.sql
```diff
@@ -1945,3 +1945,131 @@ limit 1;
...
+#
+DROP FUNCTION IF EXISTS diagnosis_noncoded;
+#
+CREATE FUNCTION diagnosis_noncoded(_encounter_id int(11), _offset int(11)) RETURNS varchar(255)
+    DETERMINISTIC
+BEGIN
+
+DECLARE ret varchar(255);
+
+select onc.value_coded into ret from obs ogr
```
mseaton, line 2013: This looks wrong - shouldn't this be selecting onc.value_text ?

mseaton (changes requested):
Overall looks good @dmdesimone .  I think I found one bug, the rest should come out in testing, but I don't see anything else obvious in the code review.  How does this perform?

configuration/pih/liquibase/sql/create_functions.sql
dmdesimone: thanks!  made the change

PR #223, opened 2021-07-20
use latestEncBetweenDates function to speed up latest vitals

dmdesimone:
use latestEncBetweenDates function to speed up latest vitals


jmbabazi:
LGTM. merging it in.

PR #226, opened 2021-07-22
Add base frontend config. Add vitals to frontend config (UHM-5648)

brandones:


mogoodrich (approved):
LGTM.

This the the config that is used if a custom config is not specified in the config-* project?

brandones:
No, it's the base config, which `config.json` is layered on top of. So we can provide config values for all PIH sites in `base-config.json`. Sites can override on a per-key basis in their `config.json`.

PR #224, opened 2021-07-20
UHM-5629 Use concept mappings when defining Program Workflows and Program Workflow States

brandones:
https://pihemr.atlassian.net/browse/UHM-5629

PR #230, opened 2021-07-23
UHM-5629 Use concept mappings when defining Program Workflows and Program Workflow States

brandones:
Same as last time, but now the HIV Treatment Status workflow is referred to by UUID, so that it isn't mixed up with a retired workflow with the same concept.

Previous PR: https://github.com/PIH/openmrs-config-pihemr/pull/224


mogoodrich (approved):
LGTM, thanks @brandones

PR #237, opened 2021-08-16
 Show last calculated EDD and gestational age at delivery time on delivery form; not working because of HFE bug UHM-5720

brandones:
It is slightly less broken than what was there before. It may or may not need to be changed again once https://pihemr.atlassian.net/browse/UHM-5720 is fixed.

### DIFF
--- a/configuration/pih/htmlforms/section-delivery.xml
+++ b/configuration/pih/htmlforms/section-delivery.xml
@@ -75,17 +75,16 @@
 
         <ifMode mode="VIEW" include="false">
             .date-one-line {
-                display:flex;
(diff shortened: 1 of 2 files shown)


mogoodrich:
Will have to dig into this a bit deeper tomorrow... is this a rollback?

brandones:
No, I just started it by rolling back my latest change. It's basically a rewrite of what was there before.

mogoodrich (approved):
I'm not following this 100%, but generally looks okay to me... I can also take a closer look once it's merged and I'm looking at HTML-5720

PR #240, opened 2021-08-20
UHM-5559 Set up relationship types in Iniz

brandones:
https://pihemr.atlassian.net/browse/UHM-5559


mogoodrich (commented):
See my comment/question on the other PR

PR #242, opened 2021-08-25
Fix the release job

brandones:
Currently, it doesn't actually deploy the maven artifact for the release version. This splits it into two so that Bamboo can run a Maven step.

See https://bamboo.pih-emr.org/build/admin/edit/editBuildTasks.action?buildKey=MIREBALAIS-CC-VR

### DIFF
--- a/ci/release-finish.sh
+++ b/ci/release-finish.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 #
-# This versions and releases a config repo.
+# This versions a config repo to the next SNAPSHOT version,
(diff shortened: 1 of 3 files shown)


mogoodrich (approved):
LGTM... I glanced at the Bamboo job as well... we actually might need the "Version and Release" to run a second mvn clean deploy, after the "release finish", to make sure that the new snapshot is deployed as well, especially if (as I think it is) the release_finish is updating PIH Core to use that version?

brandones:
@mogoodrich For that one, Bamboo will pick up its own change

Screenshot from 2021-08-25 08-17-52

PR #243, opened 2021-08-25
Remove 'name' attribute from relationship type

brandones:
Tested starting a server, no errors.

See https://github.com/mekomsolutions/openmrs-module-initializer/pull/131


brandones:
@mogoodrich @mseaton

mseaton:
@brandones - looks fine to me, but if this PR is sitting around not getting merged, my guess is it is because you don't have a ticket in PIH JIRA that is documenting this change, and which is sitting in our active sprint.  Getting the ticket created, and putting it into code review would be the best way to move this forward.

Also, from what I can tell, the Iniz PR you reference above is not yet merged:  https://github.com/mekomsolutions/openmrs-module-initializer/pull/131.  Not sure if this is a dependency of this (eg. if name is a required field on the Iniz domain until your change is made), or if these can be committed independently of each other.

brandones:
It's just a code cleanup thing, it has no consequences for users. So the kind of thing that I often just do without ticketing. This doesn't depend on the Iniz PR; `Name` is already optional.

mseaton (approved):

PR #245, opened 2021-08-31
UHM-5737 Added edge information for PMTCT history section

lnball:


mogoodrich:
Was this missing link with the edge information?

lnball:
Unfortunately I had a mess with github, so I just pushed the code out.  It didn't work (ie. show anything on the edge) but it didn't break anything.  I also undid the template in the encounterTypeConfig.js and pushed that.

If anyone wants to revisit they can see where I tried.

PR #246, opened 2021-08-31
Data kind

jmbabazi:


configuration/pih/liquibase/sql/create_functions.sql
```diff
@@ -2127,3 +2129,54 @@ RETURN ret;
...
+
+BEGIN
+    DECLARE creatorName TEXT;
+
+    SELECT CONCAT(given_name, ' ', family_name) creator_names INTO creatorName
+    FROM person_name pn
+    JOIN users u ON pn.voided = 0 AND u.retired = 0 AND pn.person_id = u.person_id
+    JOIN encounter e ON e.creator = u.user_id AND e.voided = 0 AND e.encounter_id = _encounter_id;
+
+    RETURN creatorName;
```
mogoodrich, line 2153: Each user likely has only one name, but what happens if for some reason a user has two names?

configuration/pih/liquibase/sql/create_functions.sql
```diff
@@ -2127,3 +2129,54 @@ RETURN ret;
...
+
+BEGIN
+    DECLARE creatorName TEXT;
+
+    SELECT DISTINCT(CONCAT(given_name, ' ', family_name)) creator_names INTO creatorName
+    FROM person_name pn
+    JOIN users u ON pn.voided = 0 AND u.retired = 0 AND pn.person_id = u.person_id
+    JOIN obs o ON o.creator = u.user_id AND o.voided = 0 AND o.encounter_id = _encounter_id;
+
+    RETURN creatorName;
```
mogoodrich, line 2179: Same question here as on encounter creator

configuration/pih/petl/jobs/data_kind/encounter/source.sql
```diff
@@ -0,0 +1,14 @@
+SELECT 
+    patient_id,
+    encounter_id,
+    encounter_type,
```
mogoodrich, line 4: encounter_type in the encounter table is just an int (foreign keying to the encounter_type table)... don't we want some method like ENCOUNTER_TYPE_NAME used here to resolve the actual name of the encounter_type?

configuration/pih/petl/jobs/data_kind/encounter/job.yml
```diff
@@ -0,0 +1,10 @@
+type: "sqlserver-bulk-import"
+configuration:
+  extract:
+    datasource: "mysql/openmrs.yml"
+    query:  "data_kind/encounter/source.sql"
+
+  load:
+    datasource: "sqlserver/openmrs_extractions.yml"
+    table: "datakind_encounter"
+    schema: "data_kind/encounter/target.sql"
```
mogoodrich, line 10: I'm wondering if we should think about adding a condition here to only run on JJ Dossen and Mirebalais?  

I created a ticket about adding support for this: https://pihemr.atlassian.net/browse/PETL-100

mogoodrich (commented):
Generally looks good @jmbabazi , though I added a couple questions.

configuration/pih/liquibase/sql/create_functions.sql
jmbabazi: thanks @mogoodrich , I added in a preffered = 1 condition

configuration/pih/petl/jobs/data_kind/encounter/source.sql
jmbabazi: fixed

PR #253, opened 2021-09-09
UHM-5710 - Limit program enrollment checkboxes such that only one can…

mseaton:
… be selected at a time.

### DIFF
--- a/configuration/pih/scripts/global/mch.js
+++ b/configuration/pih/scripts/global/mch.js
@@ -104,3 +104,21 @@ function setUpNextButtonForSections(currentSection) {
     window.htmlForm.submitHtmlForm();
   });
 }
+
+function setupCheckboxRadioGroups() {
+  jq(".checkbox-radio-group").each(function(groupIndex, groupElement) {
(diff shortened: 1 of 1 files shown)


lnball (approved):
LGTM

configuration/pih/scripts/global/mch.js
```diff
@@ -104,3 +104,21 @@ function setUpNextButtonForSections(currentSection) {
     window.htmlForm.submitHtmlForm();
   });
 }
+
+function setupCheckboxRadioGroups() {
+  jq(".checkbox-radio-group").each(function(groupIndex, groupElement) {
+    const options = jq(groupElement).find(".checkbox-radio-option");
+    jq(options).each(function(index1, option1) {
```
brandones, line 111: Any reason to require the `.checkbox-radio-option` classes, and not just apply this to every `input:checkbox` within the `.checkbox-radio-group`?

mseaton: Not really, probably overkill.  It's theoretically possible that a form layout could have other checkboxes nested in here that you wouldn't want affected.

PR #261, opened 2021-09-17
UHM-5781 Fix "goToNextSection" functionality 

brandones:
https://pihemr.atlassian.net/browse/UHM-5781

The code had been migrated out of the HTMLForm without changing the `&amp;` to an `&`. So whereas HFE parsed the `&amp;` into an `&` and inserted `&` into the URL bar, the script just put `&amp;` directly into the URL bar.


brandones:
I recommend reviewing the two commits separately, since the second moves code around.

mseaton (approved):
Fine with me.  Not 100% sure why this was necessary (other than the name of the file all of thi was in before), but no objections here.

mogoodrich:
Ah, right, that makes sense to be @brandones ... LGTM!

PR #264, opened 2021-09-22
Add initial consult encounter type and form for Peru

brandones:
https://github.com/PIH/openmrs-module-pihcore/pull/309
https://github.com/PIH/openmrs-module-mirebalais/pull/187


mogoodrich (approved):
LGTM... the only comment would be to confirm that we couldn't just use the "Consultation" encounter type for this (but from your description is does sound different).

mogoodrich:
Actually, note that we do have Primary Care Initial and Follow-up Encounter types already... @brandones 

(I'm fine with this as-is if we are clear that this is something different, but just wanted to point it out)

PR #268, opened 2021-09-28
UHM-5812: Rename Patient ID to EMR ID

mogoodrich:


configuration/messageproperties/messages_en.properties
```diff
@@ -3076,3 +3076,6 @@ mirebalais.cohortDefinition.diedExitFromLocationDuringPeriodSoonAfterAdmission.n
 mirebalais.cohortDefinition.diedExitFromLocationDuringPeriodSoonAfterAdmission.description = Had an Exit From Inpatient encounter at {{location}} between {{startDate}} and {{endDate}}, for which the previous disposition was Death, AND died within 48 hours of admission
 mirebalais.cohortDefinition.diedExitFromLocationDuringPeriodNotSoonAfterAdmission.name = Died in Inpatient Ward during period, at least 48h after admission
 mirebalais.cohortDefinition.diedExitFromLocationDuringPeriodNotSoonAfterAdmission.description = Had an Exit From Inpatient encounter at {{location}} between {{startDate}} and {{endDate}}, for which the previous disposition was Death, AND did not die within 48 hours of admission
+
+# override a property defined in Core Apps
```
mogoodrich, line 3080: So this overrides the property defined in Core Apps... I'm not sure if that guaranteed/defined functionality in Iniz, but it seems nice.

As @mseaton  suggested I think we could make a "Haiti-only" file if we only want to make this change for Haiti, but that might be overkill, I don't think other sites will care too much... maybe with the exception of Peru and Mexico, but then @brandones  you might just want to provide a Spanish translation?

cioan (approved):
LGTM. Thanks @mogoodrich !

configuration/messageproperties/messages_es.properties
```diff
@@ -50,7 +50,7 @@ pihcore.patientRegistration.contact.address.shortcut=(Optional) Quick search by
 
 pihcore.medicalhistory.hypertension=Hipertensión
 pihcore.medicalhistory.sti=ETS
-pihcore.acuteRheumaticFever.abbrev=ARF
```
mseaton, line 53: What's up with this change?  This doesn't look like an abbreviation.  Otherwise PR looks good.

mseaton (approved):

configuration/messageproperties/messages_es.properties
mogoodrich: I logged into Transifex and confirmed that this was a change that Miguel made through the Editor (and just got pulled in to the message.properties files when I added this code), so I am going to assume this change was correct and intentional.  fyi @brandones

PR #272, opened 2021-10-04
rewrite allpatientswithids

dmdesimone:
Rewrite allpatientswithids as it was failing in HUM prod

### DIFF
--- a/configuration/reports/reportdescriptors/dataexports/sql/allpatientswithids.sql
+++ b/configuration/reports/reportdescriptors/dataexports/sql/allpatientswithids.sql
@@ -1,70 +1,111 @@
-select 1;
-/**
 CALL initialize_global_metadata();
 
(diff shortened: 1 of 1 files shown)


dmdesimone:
@mseaton - reasons I didn't use functions:
 - phone number, names:  Yes I should use them.  I will quickly rewrite (didn't realize we had individual functions for given, family names)
  -  Address:  Since I needed to join in the person_name table any for the CDC_ID (which we don't have a function for) it didn't make sense
  - identifiers:  didn't use the function for the biometric identifier since the report really needs the DATE of the last biometric identifier, not the actual identifier.

dmdesimone:
also @mseaton we don't have a function for current age, only age_at_encounter.  

I certainly could have written functions for these missing ones but since this export isn't our highest priority right now, I just threw it together quickly.

mogoodrich (approved):

PR #278, opened 2021-10-05
Uhm 5801 mch patient changes

jmbabazi:


dmdesimone:
@jmbabazi I think you need "IF EXISTS" when creating that temp table so it doesn't fail the first time

PR #282, opened 2021-10-08
UHM-5836: Add late med pickup, late to appointment on HIV Clinical Da…

mogoodrich:
…shboard


configuration/messageproperties/messages_es.properties
```diff
@@ -3067,6 +3067,10 @@ zl.registration.patient.email.question=Ingresar correo electrónico
 pihcore.referral=Referral
 pihcore.transfer=Transfer
 
+pihcore.hivAlerts.alerts=ALERTS
```
mseaton, line 3070: Is our generally agreed best practice to make sure that all message codes are in every file, but with english text if not yet translated?  If so, this looks good.

mseaton (approved):

PR #291, opened 2021-10-29
Uhm 5762

mseaton:
This PR removes all of the PETL jobs from the config-pihemr in favor of jobs in specific implementation config projects.


configuration/pih/liquibase/sql/create_views.sql
```diff
@@ -0,0 +1,41 @@
...
+                                                                and n2.voided = 0
+                                                              order by n2.preferred desc, n2.date_created desc
+                                                              limit 1)
+         LEFT OUTER JOIN person_address a ON a.person_address_id = (select person_address_id
+                                                                    from person_address a2
+                                                                    where a2.person_id = p.person_id
+                                                                      and a2.voided = 0
+                                                                    order by a2.preferred desc, a2.date_created desc
+                                                                    limit 1)
+where p.voided = 0;
```
mseaton, line 41: By moving this to a separate SQL file that liquibase executes, rather than embedded scripts, it allows us to execute this SQL as well in our PETL process, to ensure that these views are available in restored databases used for reporting.

dmdesimone (approved):
looks good!

jmbabazi:
Thanks @mseaton, this looks good

PR #292, opened 2021-11-04
Move vitals messages into their own file

brandones:


brandones:
@mseaton The other option we should consider is simply having SES override the whole `messages_es.properties` file. It would be easier to maintain. However, SES and CES wouldn't benefit from each others' translations.

mogoodrich:
Ugh.  This seems like the right way to go about this, but it's exposing the known underlying issue that our messages.properties codes are very hap-hazard at this point, and maybe this is the point that we need to clean them up, but that's a big undertaking...

How did you determine which codes needed to be moved?  One approach would be to start standardizing on better naming conventions where the form where a message code is used is referenced directly in the form name... but that of course prohibits sharing message codes across forms... which *may* be an acceptable trade-off for knowing where codes are used?

Anyway, this may just be unhelpful brainstorming at this point, and I'd be somewhat okay with merging this in as start if it's pressing for Peru.  (I do like this better than overriding the whole messages_es file).


Happy to discuss after standup or at some other point.  We should also loop in @lnball .mAnd we should ticket this for reference.

lnball:
Someone should explain the problem to me.  It would be horrible to prohibit sharing message codes across forms.  @mogoodrich @brandones

mseaton (commented):
Before doing this, I think we should look at the new Message Source I coded up and proposed for Iniz late last week / weekend.  The PR is here:  https://github.com/mekomsolutions/openmrs-module-initializer/pull/156

Assuming it doesn't already work (which is why you are doing this), my expectation is that this will fix things such that localization actually works as one would expect - i.e. a message in messages_es_MX would override messages_es, and a message in messages_es_PE would override messages_es, and both MX and ES would fallback to es, followed by falling back to the system locale.

brandones:
Nice, that's great @mseaton . I've just checked and it does seem to work. Opened https://github.com/PIH/openmrs-config-ses/pull/46

mogoodrich:
Sounds like @lnball is voting against not being able to share message properties across forms... :)  Fine with me, @lnball  I just thought it might possibly make form translation easier because if you wanted to update/change the text for a message code you wouldn't need to worry about tracking down all the places it was used.  Sounds like this is clearly the lesser of two evils for you though.

@lnball we also stopped using Transifex in PIH-EMR while you were away because it seemed like it was becoming too much of an administrative overhead since we were at the point where 98% of the time the developer was doing the translations, but worth discussing as well.

But, coming back from that tangent... yes, I think what @mseaton suggests is the way to go (and looks like @brandones commented in agreement as well while I was writing this).

PR #296, opened 2021-11-17
UHM-5949 - modifications to the language files to harmonize them with…

mseaton:
… the english file, remove any default english translations, add in any missing codes, and harmonize on organization/formatting.


mogoodrich (commented):
Awesome, thanks... thoughts on whether it would be easy to modify the script to include the English source text in the translation files when there is no translation, to ease translation? (and also to make it easier to identifiy the right code with the i18n uuid references)

PR #297, opened 2021-11-18
UHM-5847 Add weight on patient HIV Clinical Dashboard

mseaton:


mogoodrich (approved):
LGTM, thanks @mseaton !

I assume "kg" is, in fact, the same in Kreyol, French and Spanish (@brandones  @dmdesimone )... (but regardless, having "kg" is better than nothing!)

PR #300, opened 2021-12-01
new functions for UHM-5954

dmdesimone:
new functions for UHM-5954 and a fix for the currentProgramState functions

### DIFF
--- a/configuration/pih/liquibase/sql/create_functions.sql
+++ b/configuration/pih/liquibase/sql/create_functions.sql
@@ -133,7 +133,7 @@ BEGIN
 END
 #
 /*
-This function accepts patient/person id and returns that person's age in months
+This function accepts patient/person id and returns that person's age in years
 */
 #
 DROP FUNCTION IF EXISTS current_age_in_years;
--- a/configuration/pih/liquibase/sql/create_functions.sql
+++ b/configuration/pih/liquibase/sql/create_functions.sql
@@ -859,88 +859,9 @@ BEGIN
 
 END
 #
-/*
-This function accepts a patient id, source & code of a concept and a locale
-It will return the single, latest value coded of the obs of that concept in the locale
-*/
-DROP FUNCTION IF EXISTS latest_obs_value_coded;
-#
-CREATE FUNCTION latest_obs_value_coded(_patient_id int(11), _source varchar(50), _term varchar(255), _locale varchar(50))
-    RETURNS text
-    DETERMINISTIC
-
-BEGIN
-
-    DECLARE ret text;
-
-    select      concept_name(o.value_coded, _locale) into ret
-    from        obs o
-    where       o.voided = 0
-		and o.person_id = _patient_id    		
-      	and       o.concept_id = concept_from_mapping(_source, _term)
-    order by obs_datetime desc limit 1 ;
-
-    RETURN ret;
-
-END
-#
-/*
-This function accepts a patient id, source & code of a concept
-It will return the single, latest numeric value of the obs of that concept in the locale
-*/
-#
-DROP FUNCTION IF EXISTS latest_obs_value_numeric;
-#
-CREATE FUNCTION latest_obs_value_numeric(_patient_id int(11), _source varchar(50), _term varchar(255))
-    RETURNS double
-    DETERMINISTIC
-
-BEGIN
-
-    DECLARE ret double;
-
-    select      o.value_numeric into ret
-    from        obs o
-    where       o.voided = 0
-		and o.person_id = _patient_id    		
-      	and       o.concept_id = concept_from_mapping(_source, _term)
-    order by obs_datetime desc limit 1 ;
-
-    RETURN ret;
-
-END
-#
-/*
-This function accepts a patient id, source & code of a concept
-It will return the single datetime value recorded latest for the obs of that concept in the locale
-*/
-#
-DROP FUNCTION IF EXISTS latest_obs_value_datetime;
-#
-CREATE FUNCTION latest_obs_value_datetime(_patient_id int(11), _source varchar(50), _term varchar(255))
-    RETURNS datetime
-    DETERMINISTIC
-
-BEGIN
-
-    DECLARE ret datetime;
-
-    select      o.value_datetime into ret
-    from        obs o
-    where       o.voided = 0
-	and 		o.person_id = _patient_id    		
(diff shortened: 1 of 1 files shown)


configuration/pih/liquibase/sql/create_functions.sql
```diff
@@ -133,6 +133,28 @@ BEGIN
...
 /*
+This function accepts patient/person id and returns that person's age in years
+*/
+#
+DROP FUNCTION IF EXISTS current_age_in_years;
+#
+CREATE FUNCTION current_age_in_years(
+    _person_id int)
+
+    RETURNS DOUBLE
```
mseaton, line 144: I believe this should return an INT not a DOUBLE

configuration/pih/liquibase/sql/create_functions.sql
```diff
@@ -835,6 +857,83 @@ BEGIN
...
 
+END
+#
+/*
+This function accepts a patient id, source & code of a concept and a locale
+It will return the single, latest value coded of the obs of that concept in the locale
+*/
+DROP FUNCTION IF EXISTS latest_obs_value_coded;
+#
+CREATE FUNCTION latest_obs_value_coded(_patient_id int(11), _source varchar(50), _term varchar(255), _locale varchar(50))
```
mseaton, line 868: This is fine I guess, but would be more re-usable to just have 2 functions like:

concept_name(_concept_id, _locale); -> returns String name
latest_obs_value_coded(_patient_id, _source, _term); -> returns int concept_id

and then compose these together as needed like:

concept_name(latest_obs_value_coded(_patient_id, _source, _term), _locale);

configuration/pih/liquibase/sql/create_functions.sql
```diff
@@ -835,6 +857,83 @@ BEGIN
...
+#
+DROP FUNCTION IF EXISTS latest_obs_value_numeric;
+#
+CREATE FUNCTION latest_obs_value_numeric(_patient_id int(11), _source varchar(50), _term varchar(255))
+    RETURNS double
+    DETERMINISTIC
+
+BEGIN
+
+    DECLARE ret double;
```
mseaton, line 900: Similar to my comment above, if you are going to have lots of these functions, you might want to consider having:

latest_obs(_patient_id, _source, _term) -> returns obs_id

and then:

value_numeric(_obs_id)
value_datetime(_obs_id)
value_coded(_obs_id);
value_coded_name(_obs_id);
obs_datetime(_obs_id);
etc.

This would also allow you have additional functions like these, and compose the results with the same set of obs functions:

first_obs(_patient_id, _source, _term);
latest_obs(_patient_id, _source, _term, _index_desc);
earliest_obs(_patient_id, _source, _term, _index_asc);

configuration/pih/liquibase/sql/create_functions.sql
```diff
@@ -2009,7 +2109,8 @@ BEGIN
   from patient_state ps
   inner join program_workflow_state pws on ps.state = pws.program_workflow_state_id and program_workflow_id =_program_workflow_id  
   where ps.patient_program_id = _patient_program_id
-  and ps.end_date is null;
+  and ps.end_date is null
```
mseaton, line 2112: The null end date is always an interesting question with programs.  Since all states are ended (I think) when a program enrollment is ended, this will end up showing no current program state.  But if you want to show the most recent program state for a given workflow in their enrollment, then you'll need to tweak this.  This can be a common issue if, say, you have a workflow for "Treatment Status" that have values that include terminal states like "Died".  In these cases, "Died" will never show up as the current program state.

mseaton (commented):
I've added some comments.  All of the queries generally look correct, but I've just added some suggestions of things to consider.

configuration/pih/liquibase/sql/create_functions.sql
dmdesimone: I like this approach. will rework as described.

configuration/pih/liquibase/sql/create_functions.sql
dmdesimone: This makes sense.  I didn't actually change that but will make a fix for what you describe.

dmdesimone: @mseaton, edited this function in the PR.  When you have a chance, take a look and let me know if this is what you had in mind.

configuration/pih/liquibase/sql/create_functions.sql
```diff
@@ -133,6 +133,28 @@ BEGIN
...
+DROP FUNCTION IF EXISTS current_age_in_years;
+#
+CREATE FUNCTION current_age_in_years(
+    _person_id int)
+
+    RETURNS int
+    DETERMINISTIC
+
+BEGIN
+    DECLARE currentAge DOUBLE;
```
mseaton, line 148: You changed the return value to an int correctly, but you also need this to be an int as well, no?

configuration/pih/liquibase/sql/create_functions.sql
```diff
@@ -1052,6 +1070,149 @@ offset _offset_value
...
+    where       o.obs_id = _obs_id;
+
+    RETURN ret;
+
+END
+#
+/*  
+This function accepts an obs_id and a locale
+It will return the concept name of that obs, in that locale
+*/
```
mseaton, line 1167: You should update this comment to reflect this method

mseaton: I might call this "value_coded_as_boolean"

configuration/pih/liquibase/sql/create_functions.sql
```diff
@@ -1052,6 +1070,149 @@ offset _offset_value
...
+#
+/*
+This function accepts a patient id, source & code of a question concept and source &code of an answer concept
+It will return a boolean set to 1 if that question and answer has EVER been recorded for a patient since the datetime passed in.
+Null can be passed in as the datetime if it is to be disregarded.
+*/
+#
+DROP FUNCTION IF EXISTS answerEverExists;
+#
+CREATE FUNCTION answerEverExists(_patient_id int(11), _source_question varchar(50), _term_question varchar(255), _source_answer varchar(50), _term_answer varchar(255), _begin_datetime datetime)
```
mseaton, line 1197: I like this one

mseaton (commented):
Looks great.  A few minor comments - nothing of major consequence.

PR #302, opened 2021-12-03
add function for pmtct contact ETL

dmdesimone:


configuration/pih/liquibase/sql/create_functions.sql
```diff
@@ -2180,6 +2180,33 @@ order by ps.start_date desc limit 1;
...
+
+BEGIN
+
+  DECLARE ret datetime;
+
+select max(pp.date_enrolled) into ret
+from patient_program pp 
+inner join program p on p.program_id = pp.program_id
+where pp.patient_id  = _patient_id
+and (pp.program_id = _program_id_or_uuid or p.uuid = _program_id_or_uuid)
```
mseaton, line 2204: This is kind of odd, but if it works...

mseaton (approved):

configuration/pih/liquibase/sql/create_functions.sql
mseaton: Based on how I'm seeing you use (or abuse) this in the other query, for safety I'd recommend more and more explicity functions:

1. get program_id given concept mapping
2. get workflow_id given _program_id and concept mapping
3. get state_id given _workflow_id and concept mapping
4. get latest program xyz given _patient_id and _program_id

configuration/pih/liquibase/sql/create_functions.sql
```diff
@@ -2418,7 +2445,32 @@ FROM
...
+
+BEGIN
+
+  DECLARE ret datetime;
 
+
+select min(ifnull(o.scheduled_date, o.date_activated)) into ret
+from orders o
+where o.patient_id  = _patient_id
+and o.order_reason = concept_from_mapping(_order_reason_source,_order_reason_term)
```
mseaton, line 2467: You need to check for voided here, I don't see that.

configuration/pih/liquibase/sql/create_functions.sql
mseaton: Did you make any of these suggested changes?

mseaton (changes requested):

configuration/pih/liquibase/sql/create_functions.sql
dmdesimone: oops...thanks.

configuration/pih/liquibase/sql/create_functions.sql
dmdesimone: There really isn't a need for this script to add these other functions.  You'll see in the main script, when getting the minimum patient_program_id, I already have the only database row I need.  Same with the patient_state.  We can discuss more if you think I should add additional functions for this.

mseaton (approved):

PR #308, opened 2022-01-06
UHM-6029 issues with date of test results on Add Past Labs

brandones:
https://pihemr.atlassian.net/browse/UHM-6029

Peek 2022-01-05 18-53

### DIFF
--- a/configuration/pih/htmlforms/labResults.xml
+++ b/configuration/pih/htmlforms/labResults.xml
@@ -105,16 +105,22 @@
 
             jq(function() {
                 getField('encounter-date.value').change(function(evt) {
-                    setValue('specimen-collection-date.value', evt.target.value);
                     getField('test-date.value').datepicker('option', 'minDate', evt.target.value);
                 });
-                setValue('specimen-collection-date.value', new Date());
+                getField('test-date.value').change(function(evt) {
+                    getField('encounter-date.value').datepicker('option', 'maxDate', evt.target.value);
+                })
+                const encounterDate = getDatePart(new Date(getValue('encounter-date.value'))) || new Date();
                 getField('encounter-date.value').datepicker('option', 'maxDate', new Date());
                 getField('test-date.value').datepicker('option', 'maxDate', new Date());
-                getField('test-date.value').datepicker('option', 'minDate', new Date());
+                getField('test-date.value').datepicker('option', 'minDate', encounterDate);
                 getField('viral-load.value').change(toggleViralLoadFields);
                 toggleViralLoadFields();
             });
+
+            function getDatePart(date) {
+                return new Date(date.getUTCFullYear(), date.getUTCMonth(), date.getUTCDate());
+            }
 
         </script>
     </ifMode>
@@ -176,12 +182,8 @@
                     <uimessage code="pihcore.lab.specimen_collection_date"/>
                 </label>
                 <div class="small required">
+                    <!-- The sputum collection date is stored as the encounter date. -->
                     <encounterDate id="encounter-date" default="now" />
-                    <div style="display: none;">
-                        <obs id="specimen-collection-date"
-                            conceptId="PIH:SPUTUM COLLECTION DATE"
-                            allowTime="false" />
-                    </div>
                     <div class="section-container">
                         <obs conceptId="PIH:11781" answerConceptId="PIH:YES"
                              style="checkbox" answerCode="pihcore.date_estimated"/>
@@ -193,7 +195,7 @@
                 <label for="test-date">
                     <uimessage code="pihcore.lab.test_date"/>
                 </label>
-                <div class="small required">
+                <div class="small">
                     <obs id="test-date" conceptId="PIH:DATE OF LABORATORY TEST" allowFutureDates="true" />
                 </div>
             </div>
--- a/configuration/pih/htmlforms/labResults.xml
+++ b/configuration/pih/htmlforms/labResults.xml
@@ -105,12 +105,13 @@
 
             jq(function() {
                 getField('encounter-date.value').change(function(evt) {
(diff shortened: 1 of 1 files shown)


configuration/pih/htmlforms/labResults.xml
```diff
@@ -175,19 +168,34 @@
...
+                </div>
+            </div>
+            <div>
+                <label for="specimen-collection-date">
+                    <uimessage code="pihcore.lab.specimen_collection_date"/>
+                </label>
+                <div class="small required">
+                    <obs id="specimen-collection-date"
+                         conceptId="PIH:SPUTUM COLLECTION DATE"
+                         allowTime="false" />
```
mogoodrich, line 191: It looks like you are adding a new obs here to store sputum collection date?  Previously, we had decided that specimen collection date would be tracked as the encounter date... for better or worse, as this form is in use, we can't change it without performing a data migration.  (fwiw, the idea was that the encounter date should get set to the date when the other obs in the encounter (ie the results) are clinically-relevant, which is the specimen collection data)

There's some general server-side validation that doesn't allow any of the obs value datetimes on a form to be after the encounter date, which is the source of the current error (this validation may not have been working previously which is why this is just coming up now).  We should disable the server-side validation for this, via the "allowFutureDates" feature, see: https://wiki.openmrs.org/display/docs/HTML+Form+Entry+Module+Reference#HTMLFormEntryModuleReference-%3Cobs%3E

Besides that, having JS logic to update the test date if necessary when the specimen date changes and vice versa makes sense to me,  thought  might be easier to just have a submission handler that validates that test date >= encounter date upon submission.  This is documented here, see "Defining javascript functions to call before form validation and submission":

https://wiki.openmrs.org/display/docs/HTML+Form+Entry+JavaScript+Reference

This not documented well/at all, but it's also possible to have the one-question-per-screen navigator jump back to a specific field when you display this.  I forget the details, but there are some API methods provided here, and it looks like there are some examples of usage within the form itself.

https://github.com/openmrs/openmrs-module-uicommons/blob/master/omod/src/main/webapp/resources/scripts/navigator/navigator.js#L119

mogoodrich (changes requested):
Thanks Brandon, see my comment on the ticket.

brandones:
Thanks for the thorough explanation @mogoodrich . I think I've fixed it.

Peek 2022-01-06 16-31

configuration/pih/htmlforms/labResults.xml
```diff
@@ -104,8 +104,14 @@
...
 
             jq(function() {
-                getField('specimen-collection-date.value').datepicker('option', 'maxDate', new Date());
+                getField('encounter-date.value').change(function(evt) {
+                    setValue('specimen-collection-date.value', evt.target.value);
+                    getField('test-date.value').datepicker('option', 'minDate', evt.target.value);
+                });
+                setValue('specimen-collection-date.value', new Date());
+                getField('encounter-date.value').datepicker('option', 'maxDate', new Date());
                 getField('test-date.value').datepicker('option', 'maxDate', new Date());
```
mogoodrich, line 113: Doesn't this line need to be removed?

configuration/pih/htmlforms/labResults.xml
```diff
@@ -104,8 +104,14 @@
...
+                getField('test-date.value').datepicker('option', 'minDate', new Date());
```
mogoodrich, line 114: Does this work properly during edit? ie, if you are are opening an encounter that has a both a encounter date and a test date in the past?

mogoodrich (commented):
Added a few questions, thanks @brandones !

configuration/pih/htmlforms/labResults.xml
```diff
@@ -175,19 +169,35 @@
...
+                <label for="specimen-collection-date">
+                    <uimessage code="pihcore.lab.specimen_collection_date"/>
+                </label>
+                <div class="small required">
+                    <encounterDate id="encounter-date" default="now" />
+                    <div style="display: none;">
+                        <obs id="specimen-collection-date"
+                            conceptId="PIH:SPUTUM COLLECTION DATE"
+                            allowTime="false" />
+                    </div>
```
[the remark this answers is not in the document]
dmdesimone: this works for me.  would like @lnball to weigh in also.

configuration/pih/htmlforms/labResults.xml
```diff
@@ -104,8 +104,14 @@
             }
 
             jq(function() {
-                getField('specimen-collection-date.value').datepicker('option', 'maxDate', new Date());
+                getField('encounter-date.value').change(function(evt) {
+                    setValue('specimen-collection-date.value', evt.target.value);
+                    getField('test-date.value').datepicker('option', 'minDate', evt.target.value);
+                });
+                setValue('specimen-collection-date.value', new Date());
+                getField('encounter-date.value').datepicker('option', 'maxDate', new Date());
```
brandones, line 112:
```suggestion
```

configuration/pih/htmlforms/labResults.xml
brandones: Good catch, thanks

configuration/pih/htmlforms/labResults.xml
brandones: It didn't, but it does now

configuration/pih/htmlforms/labResults.xml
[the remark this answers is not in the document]
brandones: I've changed this. I had thought I'd seen the `setValue` function in HFE working, but it now seems that it doesn't work at all, at least with date-type obs. There's no more specimen collection date obs. It's just the encounter date.

mogoodrich (approved):
LGTM, thanks @brandones !

PR #310, opened 2022-01-13
UHM-6005 - Export for user role and permission audit

mseaton:
@dmdesimone - rather than create a new export, this updates the query of an existing export.  All of the same columns are present as in the pre-existing export, they are just renamed to reflect the requirements on the ticket.  I also preserved the "Date created" column that was previously present.


dmdesimone (approved):
Look good!

PR #313, opened 2022-01-18
initial program location name function

jmbabazi:


dmdesimone (commented):
The only thing about this @jmbabazi is that I think this would fail on the 2nd statement if there were multiple enrollments on the same  patient, program, date_enrolled (which exists in the database).  

I would advocate using a "limit 1".  And you could do that in the first statement by just getting the location_name(location_id) and using:
 order by date_enrolled asc, ifnull(date_completed,'9999-12-31') asc limit 1 

(you have to account for null date_complete, which won't sort right)

dmdesimone (approved):

PR #315, opened 2022-01-24
current program location function

jmbabazi:


configuration/pih/liquibase/sql/create_functions.sql
```diff
@@ -2688,5 +2688,28 @@ BEGIN
...
+    DETERMINISTIC
+
+BEGIN
+	DECLARE currentProgramLocationName VARCHAR(255);
+
+    SELECT LOCATION_NAME(location_id) INTO currentProgramLocationName FROM patient_program
+    WHERE patient_id = _patient_id
+    AND program_id = _program_id
+    AND voided = 0
+    ORDER BY date_enrolled DESC, IFNULL(date_completed,'9999-12-31') ASC LIMIT 1;
```
mseaton, line 2710: @jmbabazi / @dmdesimone - this looks good although I'd think we'd want to order by date_completed DESC, no?  eg. if there are 2 enrollments that start on the same date, then take the one that is either still open (null date completed), or which completed most recently (later date completed).  Right?

mseaton (changes requested):

configuration/pih/liquibase/sql/create_functions.sql
dmdesimone: right... DESC

jmbabazi: fixed!

PR #319, opened 2022-01-27
UHM-6104 - Export for user role and permission audit take 2

mseaton:


mseaton:
Export looks like this, sorted by username alphabetically:

image

PR #320, opened 2022-01-27
Fix liquibase change author

brandones:


mogoodrich (commented):
Not sure if this will break the checksum on the changeset, did you test?  (I'd be fine with leaving this as it is, you could just add a  comment before if you wanted.

brandones:
I tested and it didn't cause any errors. It's idempotent, anyway.

mogoodrich:
Cool... yeah, I wasn't worried about the changeset itself running twice...  just if liquibase notices a change to changeset, it fails automatically, and I didn't know if was the case for author as well, I'm fine with merging if it worked...

mogoodrich:
Cool... yeah, I wasn't worried about the changeset itself running twice...  just if liquibase notices a change to changeset, it fails automatically, and I didn't know if was the case for author as well, I'm fine with merging if it worked...

PR #321, opened 2022-01-29
UHM-6075 - Don't show empty obs on view (showIfEmpty)

mseaton:


mogoodrich (approved):
LGTM... I noticed that the actual call to the method was comments out, but I see from your comment on the ticket that this is intentional.

PR #323, opened 2022-02-01
UHM-6099 Add HIV Test Date/Result to ObGyn form

brandones:
Peek 2022-02-01 12-00


mogoodrich (approved):
Generally defer to @lnball  on concepts and requirements, but the code looks good to me!

lnball:
Generally it looks good @brandones .  However you should check with @dmdesimone or Julia about the test result.  We already capture viral load results.  Is that different for these purposes?  What would it mean for one to be positive and another negative :(

dmdesimone:
Yeah, this was something we specifically asked for to help in reporting that Julia/Kasseem agreed that we needed.  It's a good point about discordant results, @lnball .  We can think about (maybe a different ticket?) about not allowing ART or Viral Load if the HIV Test result is negative

PR #326, opened 2022-02-08
UHM-6102 Add residency duration question

brandones:
Screenshot from 2022-02-08 10-08-00


lnball:
This "housing" question should be at the end (not beginning).  Please move it.  It's not most important.

brandones:
Was just following what was written in the ticket. It's been updated.

lnball (approved):
Looks good.  Toss it to me for testing.  Will do that as soon as code is pushed.

PR #331, opened 2022-02-11
UHM-6161: Diagnosis not voided correctly

mogoodrich:


lnball:
The liquibase looks right.  What causes this to happen?  Is it a diagnosis that's added with the post-submission and then unchecked later?

PR #336, opened 2022-02-18
UHM-6194 Remove org.openmrs.module.mirebalaisreport mappings and source

lnball:
See ticket for more information.  This is messy with updates in 2 modules and some issues.


mseaton (commented):
This looks fine if we know that there are no concepts in any packages that refer to this source.  I don't see any package updates in this commit.  If that's true, feel free to merge.

PR #342, opened 2022-03-01
Uhm 6249 add active drug concept name list function

dmdesimone:


dmdesimone:
these were merged in a previous commit

PR #343, opened 2022-03-02
UHM-6120 - Clean up SQL functions script

mseaton:


dmdesimone (approved):
Looks good!  Nice catch on the diagnosis_noncoded function.

PR #345, opened 2022-03-10
UHM-6190 Liquibase change for removing obsolete reference mappings an…

lnball:
…d sources

### DIFF
--- a/configuration/pih/liquibase/liquibase.xml
+++ b/configuration/pih/liquibase/liquibase.xml
@@ -466,4 +466,28 @@
         </sql>
     </changeSet>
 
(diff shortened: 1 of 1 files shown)


brandones (approved):
LGTM, assuming you've tested this locally without incident :)

lnball:
@brandones I did test locally on my SDK.  The changeset ran but I didn't have any of those mappings.

brandones:
Probably best if @mseaton or @mogoodrich look at it to make sure it's not subject to that subquery bug they ran into, in that case.

mseaton (approved):
Looks good to me.

PR #347, opened 2022-03-22
Configure pihemr message properties to load first so that implementat…

mseaton:
…ion-specific message properties can override them.


mogoodrich (approved):
LGTM, thanks @mseaton , though part of me kind of liked starting to break things up into different messages.properties for organizational purposes...

PR #349, opened 2022-03-29
UHM-6247: Show Past Lab Results in Lab locations

mogoodrich:


configuration/locationtags/locationtags.csv
```diff
@@ -36,4 +36,5 @@ a44acd7c-2e5d-11e9-b210-d663bd873d93,,Mental Health Location,Signifies Locations
 160a6e45-9274-4dd8-abf7-e46af8cbb41a,,Inpatient Transfer Location,
 C4BCB2BC-94E8-40DC-BB53-7384EAC6F1BF,,Program Location,Patient programs options are limited to the locations tagged with this tag.
 ae29295b-2795-11ec-9efc-0242ac110002,,Echocardiogram Location,
-278ba6a4-7570-11ec-be0e-0242ac110002,,Health Facility Location,"Signifies Locations that represent Health Facilities"
\ No newline at end of file
+278ba6a4-7570-11ec-be0e-0242ac110002,,Health Facility Location,"Signifies Locations that represent Health Facilities"
+cc104bff-6eaf-452e-a192-400d189cd82a,,Labs Component Location,"Signifies Locations where the Labs component should be enabled"
```
mogoodrich, line 40: We now can control where the "Labs" component shows up via this new location tag.

cioan (approved):
LGTM!

configuration/locationtags/locationtags.csv
mseaton: I don't love this tag name @mogoodrich .  I would think our location tags should be worded to express the types of clinical activities or workflows that happen at a location, not the software components that should be used at a location.  Is there other wording we can use here that expresses more the use cases that these locations support?

mseaton (commented):
See comment.

configuration/locationtags/locationtags.csv
mogoodrich: I wasn't in love with the name either, @mseaton , but not sure I have a better name.  It's a consequence of our two separate workflows for adding lab results... this specific requirement  was for hiding/showing this particular component at certain locations, separate from whether or not we show the different "Add Lab Results" component, which already has it's own "Lab Results Location Tag".  Hopefully once we standardize on a single component for adding lab results we can clean up the naming conventions.

PR #352, opened 2022-04-15
Move programs to CES

Joseph-Developer:


brandones:
https://github.com/PIH/openmrs-config-ces/pull/74

mogoodrich (commented):
@brandones  @Joseph-Developer  what is the intent here?  if there are CES-specific programs it does seem like they should be in CES, but these seem like more generic programs?  And, in particularly I know at least one (Mental Health) is used at other sites, and so should be globally defined (or at least individually defined in each country config).

brandones:
CES needs to change the program outcomes set for the Mental Health program, and possibly other programs. How should they go about doing that?

mogoodrich:
@brandones @Joseph-Developer the PR you issues in CES should suffice, if you rename it from "programs-ces.csv" to "programs.csv"... when building the country-specific config, any file in the country-specific package with the same name as one in config-pihemr will override the one in pihemr.

You can even keep the same constants file from pihemr (I believe).

I think this PR can be closed and I'll comment on the other one.

PR #365, opened 2022-08-01
Viral load history label translation

ljoedavessenn:
Add translation label for viral load history section

### DIFF
--- a/configuration/messageproperties/messages_en.properties
+++ b/configuration/messageproperties/messages_en.properties
@@ -3373,3 +3373,10 @@ ui.i18n.EncounterType.name.0CF4717A-479F-4349-AE6F-8602E2AA41D3=Pediatric (home)
 ui.i18n.EncounterType.name.0E7160DF-2DD1-4728-B951-641BBE4136B8=Maternal Post-partum (home)
 ui.i18n.EncounterType.name.690670E2-A0CC-452B-854D-B95E2EAB75C9=Maternal Follow-up (home)
 ui.i18n.EncounterType.name.fac9d9a2-d0bc-11ea-9995-3c6aa7c392cc=PEDS
+
+#Viral Load History
+pih.app.hiv.viralLoadHistory.title=Viral load history
+pih.app.hiv.viralLoadHistory.date=Date
+pih.app.hiv.viralLoadHistory.qualitative=Status
+pih.app.hiv.viralLoadHistory.load=Viral load (copies/ml)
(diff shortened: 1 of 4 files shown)


configuration/messageproperties/messages_en.properties
```diff
@@ -3373,3 +3373,10 @@ ui.i18n.EncounterType.name.0CF4717A-479F-4349-AE6F-8602E2AA41D3=Pediatric (home)
 ui.i18n.EncounterType.name.0E7160DF-2DD1-4728-B951-641BBE4136B8=Maternal Post-partum (home)
 ui.i18n.EncounterType.name.690670E2-A0CC-452B-854D-B95E2EAB75C9=Maternal Follow-up (home)
 ui.i18n.EncounterType.name.fac9d9a2-d0bc-11ea-9995-3c6aa7c392cc=PEDS
+
+#Viral Load History
+pih.app.hiv.viralLoadHistory.title=Viral load history
+pih.app.hiv.viralLoadHistory.date=Date
```
lnball, line 3379: This isn't wrong.  It will work.  But best practice is using and creating more general messages (ie.  zl.date and pihcore.status).  No need to add new message properties especially when there's nothing specific about viral load.

lnball:
Do you want to change to use the existing messages before merging it in.

lnball (changes requested):
See comment.  If you want to leave it, I'll approve.  I recommend using the existing messages.

configuration/messageproperties/messages_en.properties
ljoedavessenn: I'll update it accordingly. Thank you.

configuration/messageproperties/messages_en.properties
```diff
@@ -3376,7 +3376,7 @@ ui.i18n.EncounterType.name.fac9d9a2-d0bc-11ea-9995-3c6aa7c392cc=PEDS
 
 #Viral Load History
 pih.app.hiv.viralLoadHistory.title=Viral load history
-pih.app.hiv.viralLoadHistory.date=Date
-pih.app.hiv.viralLoadHistory.qualitative=Status
+pih.app.date=Date
```
lnball, line 3379: You don't need to add pih.app.status or pih.app.date.  Use "pihcore.status" and "zl.date" for your labels.

lnball (changes requested):
Small change...

lnball (approved):
Looks good

PR #366, opened 2022-08-05
UHM-6609, display previous sterilization procedures

cioan:
Hi @lnball , you just need to update whProcedures variable to point to the UUID of the sterilization procedures when those get added to the input forms. This probably needs a bit more css and html work to look nicer. Please feel free to edit it. Thanks. Here is how it looks now if the patient had any of those procedures:

Screen Shot 2022-08-05 at 11 05 50 AM

### DIFF
--- a/configuration/pih/htmlforms/section-obgyn-initial.xml
+++ b/configuration/pih/htmlforms/section-obgyn-initial.xml
@@ -157,6 +157,8 @@
     const obGynEncounterType= 'd83e98fd-dc7b-420f-aa3f-36f648b4483d';
     const typeOfVisit = '164181AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA';
     const intakeVisit = '164180AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA';
+    const whProcedurePerformed="1651AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA";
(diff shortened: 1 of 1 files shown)


configuration/pih/htmlforms/section-obgyn-initial.xml
```diff
@@ -157,6 +157,8 @@
     const obGynEncounterType= 'd83e98fd-dc7b-420f-aa3f-36f648b4483d';
     const typeOfVisit = '164181AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA';
     const intakeVisit = '164180AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA';
+    const whProcedurePerformed="1651AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA";
+    const whProcedures="57755670-c183-494b-bd5e-723e0978c3e3";
```
lnball, line 161: I like the list AND adding dates is great.  Shouldn't this be the other convset?

const whSterilizationProcedures = "46cbbdd9-9cc2-42c6-8ecb-4dbe6ea01354";

image

lnball (changes requested):
See comment.

configuration/pih/htmlforms/section-obgyn-initial.xml
lnball: AND nice work @cioan !

PR #368, opened 2022-08-11
UHM-6643: Lab OWA Assessments and Tweaks

mogoodrich:


configuration/globalproperties/gp.xml
```diff
@@ -363,10 +363,15 @@
             <property>labworkflowowa.estimatedCollectionDateAnswer</property>
             <value>${concept.yes.uuid}</value>
         </globalProperty>
-        <globalProperty>
+        <!--<globalProperty>  NO LONGER USED, CAN BE REMOVED FROM EXISTING SERVERS
             <property>labworkflowowa.labResultsToDisplayConceptSet</property>
-            <!-- Concept: Laboratory concept set -->
+            &lt;!&ndash; Concept: Laboratory concept set &ndash;&gt;
             <value>d50d46db-3b81-4a5a-b3c6-01857aa9c520</value>
+        </globalProperty>-->
```
mogoodrich, line 370: I didn't say a way in the Iniz specs to remove a GP entirely, does anyone know if this is possible? (Otherwise, I can add a liquibase changeset to do so)

mseaton: Yeah, I would remove the above global property declaration completely, rather than comment it out, and then add a liquibase changeset to that domain to remove it.

mseaton (changes requested):

mogoodrich:
Updated the script to remove the labworkflowowa.labResultsToDisplayConceptSet GP entirely via liquibase.  Merging this in.

PR #369, opened 2022-08-19
Create update_registration_encounter_datetimes,sql

dmdesimone:


mseaton (commented):
The file itself looks fine @dmdesimone .  I think the right location for this file is in the openmrs-config-zl repository (since this is a ZL-specific migration), and I think we should put it in a new folder at "configuration/pih/migrations".  We should add a README.md to this folder basically indicating that this folder contains data migration scripts that were not executed by liquibase but as one-off updates.  And then we should add comments to the top of this SQL file that explains what it does and why, and points back to the source ticket.  Doable?

dmdesimone:
Will do, @mseaton !  thanks.

PR #370, opened 2022-08-22
UHM-6626 For Cosmin

lnball:
Thanks @cioan !


cioan:
@lnball , I think I got it working. The problem is that on this section-obgyn-initial form there are two calculated-edd-and-gestational divs and the jQuery gets confused when trying to find by ID. I changed to find by class and now the calculated values show up. But, I think there are bunch of other forms that need to be updated, or refactor the mch.js function to factor in the the ID of the DIV.

cioan (approved):
Looks good, @lnball ! Thanks!

PR #372, opened 2022-09-07
UHM-6532, include programstatus widget into the OB/GYN form

cioan:
Here is the uiInclude tag that adds the programstatus widget to the form:

Screen Shot 2022-09-07 at 9 48 02 AM


lnball:
I like the approach.  How does it change the css when all the class+id are within the div?  It would be good to play with the browser css tool.  Can you move the code OUTSIDE of the section or section-container?

PR #373, opened 2022-09-09
refactor to reduce run time

dmdesimone:


mseaton (changes requested):
@dmdesimone I brought this question up with @jmbabazi before you joined the call, but I think we need to resolve it before I review this:  I don't think this is the right file/query.  There is another j9CaseRegistration.sql file in the openmrs-config-zl project.  I don't know why the query exists in both, do you?  Do we do j9 outside of Haiti?  Can we delete this SQL out of the config-pihemr, and consolidate changes into the one in config-zl?  Then apply your changes, and I can review...

dmdesimone:
removed  j9CaseRegistration.sql from this repo.  Will reopen PR in openmrs-config-zl, where this file belons

PR #374, opened 2022-09-21
UHM-6501 - Track user logins in OpenMRS

mseaton:


configuration/log4j2.xml
```diff
@@ -0,0 +1,82 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!--
```
mogoodrich, line 2: Great to have this file out of the code and into config... is this all the same as what is in the code except for the MFA stuff?

mogoodrich (approved):
One comment, but LGTM!

configuration/log4j2.xml
mseaton: You can review better below now @mogoodrich - I committed the base version to the master branch and then merged up master and fixed the conflicts.  So this now shows a diff from what is in core.

PR #385, opened 2022-10-26
UHM-6758 - Suppress warnings from Hibernate.

mseaton:


mogoodrich (approved):
LGTM, though maybe add a comment here that references the ticket and/or the issue, otherwise we'll never remember why we did this?

mseaton:
> maybe add a comment here that references the ticket and/or the issue

Done, merging

PR #386, opened 2022-11-01
add creation of petl_load_times

dmdesimone:


configuration/pih/liquibase/liquibase.xml
```diff
@@ -643,5 +643,19 @@
         <comment>Remove old authentication events following bug fixes</comment>
         <delete tableName="authentication_event_log"/>
     </changeSet>
-
+        
+    <changeSet id="20221101-initial-create-of-petl-load-times-table" author="ddesimone">
+        <preConditions onFail="MARK_RAN">
+            <sqlCheck expectedResult="0">
+                select table_exists("petl_load_times");
```
mseaton, line 650: Better to follow existing standard and conventions here.  See where I add the authentication_event_log table.  Something like:

```<not><tableExists tableName="petl_load_times"/></not>```

File: configuration/pih/liquibase/liquibase.xml [hunk 2]
```diff
@@ -643,5 +643,19 @@
...
+            </sqlCheck>
+        </preConditions>            
+        <sql>
+                drop table if exists petl_load_times;
+                create table petl_load_times
+                (loaded_domain	        varchar(50),		
+                last_loaded_time	datetime);
+        </sql>
```
mseaton, line 658: Again, better to use the liquibase constructs for this, as done throughout this liquibase.xml file.  First, no need (and not desireable) to have the `drop table if exists...` statement.  Your precondition already guarantees that it doesn't exist.  Second, in order to create the table, use the liquibase tags for this (again, see where I add the authentication event log table).  Something like:

```
        <createTable tableName="petl_load_times">
            <column name="loaded_domain" type="varchar(50)"/>
            <column name="last_loaded_time" type="datetime"/>
        </createTable>
```

mseaton (changes requested):

PR #387, opened 2022-11-03
Uhm 6820

cioan:
@mogoodrich , this small change below seems to fix the problem I am describing in https://pihemr.atlassian.net/browse/UHM-6820

I am curious why you initially left out the last input field (Administration instructions) from running through the validation function?

### DIFF
--- a/configuration/pih/htmlforms/dispensing.xml
+++ b/configuration/pih/htmlforms/dispensing.xml
@@ -5,13 +5,14 @@
             var numErrors = 0;
             var numDispensed = 0;
             var scrolled = false;
-
+console.log("getBeforeValidation");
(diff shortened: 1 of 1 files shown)


mogoodrich (approved):
LGTM,  @cioan ... to your question, I'm not sure... I moved this file over from the dispensing module in 2020 so that it where the commit history comes from, I don't remember if I was the one that actually wrote this back in the day...

My quess would be that the reasoning was becaues perhaps administration instructions was thought not to be required at some point?

cioan:
Thanks @mogoodrich ! Sorry, I thought you originally wrote this validation function.

mogoodrich:
No worries @cioan !

PR #388, opened 2022-11-07
LIB-166, add out-of-stock checkbox

cioan:
Hi @lnball , I was able to get the new out-of-stock checkboxes working on the humci instance. I think it just just a matter of copying these changes below to the Liberia dispensing form. Please let me know if you want me to do that. Thanks!

### DIFF
--- a/configuration/pih/htmlforms/dispensing.xml
+++ b/configuration/pih/htmlforms/dispensing.xml
@@ -7,12 +7,15 @@
             var scrolled = false;
 
             var requiredAndEmpty = function(jqEl) {
-                // medication-instructions is optional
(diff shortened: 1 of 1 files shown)


lnball:
Go for it @cioan .  I just checked in a new mds.  You can try it.  I will also build a new one from scratch now

configuration/pih/htmlforms/dispensing.xml
```diff
@@ -39,6 +42,7 @@
                 var anyErrors = false;
                 if (numFilledFields > 0) {
                     jq(this).find('input, select').each(function () {
+                      debugger
```
lnball, line 45: Remove debugger ;)

PR #392, opened 2023-01-19
UHM-6871: Get the new O3 ordering into the build

mogoodrich:
This adds the proper configuration for the Medications app (order baskset) based on the PIH EMR metadata.
Note the comment I added about not wanting to us the Drug Order Encounter long-term... hopefully we will have some sort of encounter context feature set up before we need to go live with this.

### DIFF
--- a/configuration/frontend/base-config.json
+++ b/configuration/frontend/base-config.json
@@ -31,5 +31,17 @@
       "generalPatientNoteUuid": "3cd9d956-26fe-102b-80cb-0017a47871b2",
       "midUpperArmCircumferenceUuid": "4326b04b-3158-417a-bb8d-ad022295b0f4"
     }
+  },
+  "@openmrs/esm-patient-medications-app": {
+    "durationUnitsConcept": "3cd706b8-26fe-102b-80cb-0017a47871b2",
+    "daysDurationUnit": {
+      "uuid": "3cd706b8-26fe-102b-80cb-0017a47871b2"
+    },
+    "careSettingUuid": "6f0c9a92-6f24-11e3-af88-005056821db0",
+    "quantityUnitsUuid": "162395AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
+    "clinicianEncounterRole": "c458d78e-8374-4767-ad58-9f8fe276e01c",
+    "drugOrderTypeUUID": "131168f4-15f5-102d-96e4-000c29c2a5d7",
+    "drugOrderEncounterType": "0b242b71-5b60-11eb-8f5a-0242ac110002",
+    "_comment": "TODO: drug order encounter type refers to Drug Order Documentation encounter; we likely will need another encounter type, or, better yet, none at all"
   }
 }


configuration/frontend/base-config.json
```diff
@@ -31,5 +31,17 @@
...
+    "durationUnitsConcept": "3cd706b8-26fe-102b-80cb-0017a47871b2",
+    "daysDurationUnit": {
+      "uuid": "3cd706b8-26fe-102b-80cb-0017a47871b2"
+    },
+    "careSettingUuid": "6f0c9a92-6f24-11e3-af88-005056821db0",
+    "quantityUnitsUuid": "162395AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
+    "clinicianEncounterRole": "c458d78e-8374-4767-ad58-9f8fe276e01c",
+    "drugOrderTypeUUID": "131168f4-15f5-102d-96e4-000c29c2a5d7",
+    "drugOrderEncounterType": "0b242b71-5b60-11eb-8f5a-0242ac110002",
+    "_comment": "TODO: drug order encounter type refers to Drug Order Documentation encounter; we likely will need another encounter type, or, better yet, none at all"
```
mogoodrich, line 45: I think "Drug Order Documentation" encounter was intended for legacy orders imported from the HIV EMR.  We don't have a vanilla "Drug Order" encounter, but I didn't want to create one because, ideally, the idea of an encounter context that allows orders to be associated with different encounter types will be in place before we go live with this.  The current encounter type just allows us to continue testing the order basket functionality in the short-term.

cioan (approved):
Thanks @mogoodrich !

configuration/frontend/base-config.json
```diff
@@ -31,5 +31,17 @@
       "generalPatientNoteUuid": "3cd9d956-26fe-102b-80cb-0017a47871b2",
       "midUpperArmCircumferenceUuid": "4326b04b-3158-417a-bb8d-ad022295b0f4"
     }
+  },
+  "@openmrs/esm-patient-medications-app": {
+    "durationUnitsConcept": "3cd706b8-26fe-102b-80cb-0017a47871b2",
```
mseaton, line 36: What does "durationUnitsConcept" represent?  What does it need configuration client-side in the first place, and are you sure this is the right value?  There is already a global property set (which we do in config-pihemr) for order.durationUnitsConceptUuid.  Is this something different?  The value we are setting for the GP is 19cd6ff8-052b-46e7-96ef-91fccede9565, which is different from what you have configured here...

configuration/frontend/base-config.json
```diff
@@ -31,5 +31,17 @@
...
+    "daysDurationUnit": {
+      "uuid": "3cd706b8-26fe-102b-80cb-0017a47871b2"
```
mseaton, line 38: This is another example of a configuration property that shouldn't need to be set/maintained here, as elsewhere in OpenMRS the various duration units are mapped to standard SNOMED-CT duration codes that can be interpreted (see BaseDosingInstructions.java and Duration.java in OpenMRS core).  This may be outside the scope of this PR, and this may well be the correct configuration, but we should challenge whether this should need to be configured here.

configuration/frontend/base-config.json
```diff
@@ -31,5 +31,17 @@
...
+    },
+    "careSettingUuid": "6f0c9a92-6f24-11e3-af88-005056821db0",
```
mseaton, line 40: This is fine I guess - we are just saying, always use "OUTPATIENT" care setting?

configuration/frontend/base-config.json
```diff
@@ -31,5 +31,17 @@
...
+    "quantityUnitsUuid": "162395AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
```
mseaton, line 41: On humci, this is the concept uuid for "Bag".  So I believe this is the incorrect uuid. Which also adds evidence that we should not need to configure this here at all.  I'm also not sure what "quantityUnits" refers to.  There are global properties (and associated services) for core that are the following:  `order.drugDosingUnitsConceptUuid` and `order.drugDispensingUnitsConceptUuid`.  Can we not just get this from the core configuration, using the one(s) that are most appropriate (assuming this is intending to represent one of these?)

configuration/frontend/base-config.json
```diff
@@ -31,5 +31,17 @@
...
+    "clinicianEncounterRole": "c458d78e-8374-4767-ad58-9f8fe276e01c",
```
mseaton, line 42: Could use a constant for this if desired:  ${encounterRole.orderingProvider.uuid}

configuration/frontend/base-config.json
```diff
@@ -31,5 +31,17 @@
...
+    "drugOrderTypeUUID": "131168f4-15f5-102d-96e4-000c29c2a5d7",
```
mseaton, line 43: This is hard-coded into OpenMRS core, wondering if we really need it to be configurable.  Or at the very least, it should be allowed to not be configured, and should default to the value from core.  See OrderType.DRUG_ORDER_TYPE_UUID

configuration/frontend/base-config.json
mseaton: Agree this is fine to use for now, hopefully we can remove the need for doing this asap.

mogoodrich:
Ugh, thanks @mseaton. I thought I double-checked but it sounds like I might have messed up some of the uuids, I will double-check.  Regarding the fact that the configuration even needs to happen here for some of these data points I can start a Talk post.

configuration/frontend/base-config.json
mogoodrich: You are right, this is definitely wrong, because it looks like the same one used for days before... I had to do some cut-and-pasting of this whole config at one point, I probably took an old version I was using.

mogoodrich:
Closing this out without merging until we get the issues I've referenced on the related ticket fixed.

PR #394, opened 2023-01-25
UHM-6871: Get the new O3 ordering into the build

mogoodrich:


configuration/frontend/base-config.json
```diff
@@ -31,5 +31,13 @@
       "generalPatientNoteUuid": "3cd9d956-26fe-102b-80cb-0017a47871b2",
       "midUpperArmCircumferenceUuid": "4326b04b-3158-417a-bb8d-ad022295b0f4"
     }
+  },
+  "@openmrs/esm-patient-medications-app": {
+    "daysDurationUnit": {
```
mogoodrich, line 36: this is just because it sets the "days" as the default duration unit and needs to know the uuid of that concept

cioan (approved):
Looks good, @mogoodrich !

configuration/frontend/base-config.json
mseaton: @mogoodrich This is fine for now I guess, but my point before was that OpenMRS ordering seems to require that all duration units are mapped to the official SNOMED-CT duration codes.  "Days" is 258703001 in SNOMED-CT, from what I gather.  If you look at any of our PIH EMR systems, you'll see a mapping of SNOMED CT:258703001 on the concept with this uuid.  And if you look at OpenMRS core, you'll see this is hard-coded as a constant in the Duration class since 1.10 (https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/Duration.java#L37).  So ultimately, I don't see why the ESM should have us configure something that OpenMRS core already knows how to look up reliably.

mogoodrich: Make sense, I'll write up a ticket about this, thanks @mseaton

PR #396, opened 2023-01-31
uhm-6757 enroll missed J9 patients

dmdesimone:


configuration/pih/liquibase/liquibase.xml
```diff
@@ -643,5 +643,53 @@
...
+			and not EXISTS 	
+				(select 1
+				from patient_state ps 
+				where ps.patient_program_id = pp.patient_program_id 
+				and (ps.end_date is null or ps.end_date > o.obs_datetime))	-- any state that overlaps with the mother's group obs
+			group by o.person_id , pp.patient_program_id;
+			
+			
+			-- add those patients to J9 workflow by adding patient_state rows for them with state = prenatal group
+			insert into patient_state (patient_program_id , state, start_date, creator, date_created,voided, uuid)
```
mogoodrich, line 689: Are we doing any validation to make sure that the patient state start date falls within the program that we chose?

mogoodrich (commented):
@dmdesimone  see my comment... do we do anything to assure that the patient state start date falls within the program start and end dates?

configuration/pih/liquibase/liquibase.xml
```diff
@@ -651,6 +651,7 @@
 				inner join patient_program pp on pp.patient_id  = o.person_id 
 					and pp.program_id = (select program_id from program where uuid = '41a2715e-8a14-11e8-9a94-a6cf71072f73') and pp.date_completed is null
 				where o.concept_id = concept_from_mapping('PIH',11665)
+				and o.obs_datetime >= pp.date_enrolled
```
mogoodrich, line 654: so we do nothing if the obs datetime is before the date enrolled (as opposed to setting the state, but on the enrollment date instead of the obs date) (maybe this is a case that never really occurs...?)  @dmdesimone

dmdesimone: Well that SHOULD never happen because the same encounter that contains the "mothers group" obs should enroll them if they are unenrolled.  And in general, I just wanted to be a little more conservative and only update data in cases where we could be 100% sure what should be done.

dmdesimone: @mogoodrich ^^

PR #398, opened 2023-02-03
UHM-6839: Set "Reason for Change" when substituting drugs

mogoodrich:
New configuration for substitution questions answered provided by PIH EMR

### DIFF
--- a/configuration/frontend/base-config.json
+++ b/configuration/frontend/base-config.json
@@ -18,6 +18,14 @@
     "Brand color #2": "#FFC166",
     "Brand color #3": "#FFD699"
   },
+  "@openmrs/esm-dispensing-app": {
+    "substitutionReason": {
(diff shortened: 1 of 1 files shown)


mseaton (approved):
LGTM, assuming these are the right uuids.  Note, I'm not a fan of configuration by uuid in general, I'd rather we be able to configure by mapping or something a little more clearly representative of the concept semantically

lnball:
LGTM.  uuids are correct but as Mike points out -- hard to read.

mogoodrich:
Yeah, agreed re:uuid... those this seems to be the general configuration pattern in O3... let me dig a little deeper and see if there's any utility methods to fetch by mappings, and, if not, ticket in the overall O3 project to add them.

mogoodrich:
Started a thread here about the best way to search/reference valueset here: https://talk.openmrs.org/t/searching-for-valuesets/38480

Going to merge this in with the uuid solution for now to keep things moving forward.

PR #399, opened 2023-02-13
UHM-7031 File all_medication_prescribed

jmbabazi:


configuration/reports/reportdescriptors/dataexports/sql/all_medication_prescribed.sql
```diff
@@ -0,0 +1,103 @@
...
+order_date_activated
+)
+select 
+encounter_id,
+patient_id,
+order_id,
+concept_id,
+date(date_created),
+date(date_activated)
+from orders  where voided = 0;
```
dmdesimone, line 51: I think you need to filter for drug orders here.  Otherwise, you'll get lab, radiology, pathology etc... orders.  (also it will run much faster)

PR #400, opened 2023-02-14
UHM-7022: unresolved label editing Prenatal Home encounter

mogoodrich:


configuration/pih/htmlforms/section-mch-referral.xml
```diff
@@ -102,7 +102,7 @@
             <p class="radio">
                 <strong>
                     <label>
-                        CommCare&amp;nbsp;<uimessage code="mirebalais.login.username"/>
+                        CommCare&amp;nbsp;<uimessage code="pihcore.username"/>
```
lnball, line 105: Why not use "emr.user.username" which was already in the message properties?

It's a simple change.  If you want to use the new codes, go for it.

mogoodrich: Nice catch, I may have missed that, let me check.

mseaton: You can also use something from the authenticationui module.

mogoodrich: Not sure how I missed emr.user.username, I thought I had searched thoroughly... switched to that...

lnball:
LGTM.  Simple ;)

PR #405, opened 2023-03-06
LIB-182: Registration: Add Homeless Checkbox

mogoodrich:


mseaton (commented):
See my other review.  I think we should reconsider this decision to model as a person attribute.  Otherwise, this code itself is fine.

PR #407, opened 2023-03-15
LIB-193 - Add CHW fields to Registration form

mseaton:


lnball (approved):
Great

configuration/messageproperties/messages_en.properties
```diff
@@ -2533,6 +2533,19 @@ zl.registration.patient.biometrics.fingerprints.rightIndexLabel=Right Index Fing
 
 zl.registration.patient.additionalIdentifiers=Additional Identifiers
 
+pih.registration.patientSupport.section.label=Patient Support
+pih.registration.patientSupport.label=Patient Support
+pih.registration.patientSupport.question=Patient Support
+pih.registration.patientSupport.type.label=Who is following the patient?
+pih.registration.following.type.chw=Community health worker
```
lnball, line 2540: @mseaton These don't need to add every label.  It's fine to re-use existing codes.  ID, name, phone contact, CHW are likely already in the file and should be reused.  And these don't need to be these long complex codes "pih.registration.following.type.chw.  More flexible to call it pih.chw or pih.role.chw.

lnball (changes requested):
Reuse...

configuration/messageproperties/messages_en.properties
mseaton: I go back and forth on this.  For now, I've done it this way because it keeps it consistent with all of the other registration forms and calf configurations.

mseaton: I'm going to merge as is, because I've already tested it like this and for the reasons I listed, but I'd support cleaning all this up @lnball

PR #409, opened 2023-03-30
UHM-7120 Create javascript functions that enable querying a patient's…

mseaton:
… conditions

### DIFF
--- a/configuration/pih/scripts/global/htmlFormFunctions.js
+++ b/configuration/pih/scripts/global/htmlFormFunctions.js
@@ -1,0 +1,33 @@
+/**
+ * Adds javascript functions for use within html forms
+ * Requires jQuery and moment js
+ */
+class HtmlFormFunctions {
+
+    constructor(pihemr, patientUuid, encounterUuid, encounterDateStr) {
+        this.pihemr = pihemr;
+        this.jq = pihemr.getJq();
+        this.patientUuid = patientUuid;
+        this.encounterUuid = encounterUuid;
+        this.encounterDateStr = encounterDateStr;
+    }
+
+    getPatientUuid() {
+        return this.patientUuid;
+    }
+
+    getEncounterUuid() {
+        return this.encounterUuid;
+    }
+
+    getEncounterDateYmd() {
+        let encDate = jq('#encounterDate').find('input[type="hidden"]').val();
+        if (!encDate) {
+            encDate = this.encounterDateStr;
+        }
+        if (!encDate) {
+            encDate = moment().format('YYYY-MM-DD');
+        }
+        return encDate;
+    }
+}
--- a/configuration/pih/scripts/global/pihServices.js
+++ b/configuration/pih/scripts/global/pihServices.js
@@ -1,0 +1,128 @@
+/**
+ * Adds javascript functions to the pihemr
+ * Must be instantiated with jQuery instance
+ */
+class PihServices {
+
+    constructor(jq) {
+        this.jq = jq;
+    }
+
+    getJq() {
+        return this.jq;
+    }
+
+    getContextPath() {
+        return window.location.href.split('/')[3];
+    }
+
+    getApiBaseUrl() {
+        return "/" + this.getContextPath() + "/ws/rest/v1";
+    }
+
+    getResourceUrl(resource) {
+        return this.getApiBaseUrl() + "/" + resource;
+    }
+
+    /**
+     * Returns a promise of concepts that match the passed conceptReferences, with the given representation
+     * @param conceptReferences a comma-delimited string of uuids or mappings in source:code format
+     * @param representation the object representation to return
+     * @returns {Promise<*>}
+     */
+    async getConceptsByReferences(conceptReferences, representation) {
+        let data = await jq.getJSON(this.getResourceUrl("concept"), {
+            references: conceptReferences,
+            v: representation
+        });
+        if (data.results) {
+            return data.results;
+        }
+        return [];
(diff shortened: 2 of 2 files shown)


configuration/pih/scripts/global/htmlFormFunctions.js
```diff
@@ -0,0 +1,33 @@
+/**
+ * Adds javascript functions for use within html forms
+ * Requires jQuery and moment js
+ */
+class HtmlFormFunctions {
```
mogoodrich, line 5: Is this mean to replace/be an improvement to stuff currently offered by JS code within the HFE module itself?

mogoodrich (commented):
Not to rabbit-hole this, but this seems like the beginning of some better-organized global JS compared to the existing stand-alone scripts like expandable.js, mch.js, nextAndSubmutButtions.js... etc, 

Not saying we need to migrate all that stuff here as part of this commit, but if we think this provides a better approach let's document that and depreciate the other stuff so in the future we try to continue following this pattern (if that's what we want).

mseaton:
> Not to rabbit-hole this, but this seems like the beginning of some better-organized global JS compared to the existing stand-alone scripts like expandable.js, mch.js, nextAndSubmutButtions.js... etc,

Possibly yes.  Let's keep that scope separate.  For now, I really just wanted to you look over this JS and consider whether this is in fact an improvement and a direction we might want to go for how this is organized.

configuration/pih/scripts/global/htmlFormFunctions.js
mseaton: Such as what @mogoodrich ?

mogoodrich: See the getField, getValue and setValue as defined here: https://wiki.openmrs.org/display/docs/HTML+Form+Entry+JavaScript+Reference

(which admittedly would't handle patientUuid and encounterUuid, I believe)

configuration/pih/scripts/global/htmlFormFunctions.js
```diff
@@ -0,0 +1,33 @@
...
+
+    getPatientUuid() {
+        return this.patientUuid;
+    }
+
+    getEncounterUuid() {
+        return this.encounterUuid;
+    }
+
+    getEncounterDateYmd() {
```
mogoodrich, line 23: Actually, referring to the above, this may fail in some cases (though admittedly this is not not well documented).  The "encounterDate" element only had an id of "encounterDate" if no specific id parameter is given:

https://github.com/openmrs/openmrs-module-htmlformentry/blob/master/api/src/main/java/org/openmrs/module/htmlformentry/element/EncounterDetailSubmissionElement.java#L563

And specifically, the property accessor stuff theoretically would be updated if we modified the structure of the encounter date widget behind the scenes.

configuration/pih/scripts/global/pihServices.js
```diff
@@ -0,0 +1,128 @@
+/**
+ * Adds javascript functions to the pihemr
+ * Must be instantiated with jQuery instance
+ */
+class PihServices {
```
mogoodrich, line 5: Compared to the above, I don't think we have anything in HTML Form Entry that provides utility methods for interacting with OpenMRS RESTfully, so it would likely make sense to add these to HFE directly, unless there is something specific here to PIH I missed?

mogoodrich (commented):
See my updated comments... I think my gut would be not to add htmlFormFunctions but update the field accessors provided by HFE as need. and add the pihServices.js to HFE and something generic call "restServices" unless there is something PIH-specific I'm missing.

Thanks for working on this, would be good to get away from all the ad hoc stuff we are doing in our PIH-specific JS code.

configuration/pih/scripts/global/htmlFormFunctions.js
mseaton: I believe those functions only work in edit/enter modes, right @mogoodrich ?  I'm fine moving away from our custom functions to ones that are in HFE, but the fact is that our HTML Forms are littered with code that gets encounter date in all sorts of custom and different ways, and often doesn't get it using these hfe functions.  So we'd want to have tickets specifically created to thoroughly review particular use cases and handle them appropriately.

configuration/pih/scripts/global/htmlFormFunctions.js
mseaton: Yes, it may fail in some cases, which is why the function is in the pihemr config and we can use it in places where we know it will work (eg. have standards for what ids we place on our form elements).  I see no suitable features that exist in HFE already that will meet the same needs that these provide.

configuration/pih/scripts/global/pihServices.js
mseaton: Well, there is also nothing specific to HFE about wanting to access REST services.  So why bundle something with HFE that could have broader uses?  I could see the REST module providing javascript library resources - that would make a lot of sense.  But that's outside the scope of what I'm doing here.

configuration/pih/scripts/global/htmlFormFunctions.js
mogoodrich: Ah, it's possible the don't work in view mode, fair enough.... 

Agree that our code is likely littered with code that gets dates in different ways, and probably doesn't use the HFE functions... not saying we need to clean this all up now, just didn't want to start propagating another method... but, yeah, if those getter/setters don't work in View mode and we need something new anyway, better to set up a standard...

PR #410, opened 2023-04-20
UHM-7160: Close Stale Visits error on startup

mogoodrich:


mseaton (approved):
LGTM.  I assume there are no corresponding rows in the scheduler_task_config_property table that would cause this to fail with a constraint issue?

mogoodrich:
Added line to delete any related entries in scheduler_task_config_property table, thanks @mseaton , merging but post-merge review welcome.

PR #411, opened 2023-04-25
LIB-214 - Update ncd and mental health forms to use javascript/css fo…

mseaton:
…r conditional condition-based logic rather than velocity expressions.

### DIFF
--- a/configuration/pih/scripts/global/htmlFormFunctions.js
+++ b/configuration/pih/scripts/global/htmlFormFunctions.js
@@ -30,4 +30,25 @@ class HtmlFormFunctions {
         }
         return encDate;
     }
+
+    async showSectionBasedOnCondition(conceptOrSet, sectionSelector, showIfActive, showIfInactive, showIfInEncounter) {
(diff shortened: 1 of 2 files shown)


configuration/pih/scripts/global/htmlFormFunctions.js
```diff
@@ -30,4 +30,25 @@ class HtmlFormFunctions {
...
+                if (hfe.pihemr.isConditionActiveOnDate(condition, encDateYmd)) {
+                    foundActiveAtEncounter = true;
+                }
+            });
+            if ((foundActiveAtEncounter && showIfActive) || (!foundActiveAtEncounter && showIfInactive) || (foundInEncounter && showIfInEncounter)) {
+                jq(sectionSelector).show();
+            }
+        });
+    }
 }
```
mseaton, line 54: Moved this here from the liberia adult ncd intake form and made it more generic, so that it could be used for a number of forms and use cases with condition list

mogoodrich (approved):
LGTM, thanks @mseaton !

PR #419, opened 2023-05-17
UHM-7182: Add actual time to timestamps in lab module exports

mogoodrich:


dmdesimone (commented):
Looks good and nice catch on result_entry_date.  Please see additional comment in ticket UHM-7182 regarding the additional change in ZL ETL that needs to occur.

PR #432, opened 2023-08-17
UHM-7344: Upgrade to Tomcat 9

mogoodrich:


configuration/globalproperties/gp.xml
```diff
@@ -404,5 +404,11 @@
             <value>configuration/frontend</value>
         </globalProperty>
 
+        <!-- OWA -->
+        <globalProperty>
+            <property>owa.appFolderPath</property>
+            <value>/home/tomcat/.OpenMRS/owa</value>
+        </globalProperty>
```
mogoodrich, line 411: There was no previous value for this, because the OWA module automatically sets the GP based on the application data directory, but doesn't update it if that changes.

mseaton (approved):

PR #435, opened 2023-09-12
UHM-7936: Custom O3 Header for SL

mogoodrich:


configuration/frontend/base-config.json
```diff
@@ -1,7 +1,7 @@
 {
   "@openmrs/esm-api": {
     "redirectAuthFailure": {
-      "url": "/pihcore/login.htm"
+      "url": "${openmrsBase}/authenticationui/login/login.page"
```
mseaton, line 4: I think this should work, but another approach would be to redirect to something like "/index.htm" and let the authentication module take care of the necessary redirection.  Basically, if you point to any URL that is served by OpenMRS but which is _not_ whitelisted, then you should be redirected to the appropriate login page if you are not yet authenticated.  In the current case, that will be the page you list above, though I suppose it might not always be...

configuration/frontend/base-config.json
```diff
@@ -55,10 +55,15 @@
...
       "src": "/openmrs/spa/site/pih-logo.png",
-      "alt": "Partners in Health"
+      "alt": "Partners In Health"
     },
     "links": {
       "loginSuccess": "${openmrsBase}/pihcore/home.page"
+    },
+    "provider": {
+      "type": "oauth2",
+      "loginUrl": "${openmrsBase}/authenticationui/login/login.page",
```
mseaton, line 65: Same comment as above, though I agree this looks more sensible.

configuration/frontend/base-config.json
```diff
@@ -55,10 +55,15 @@
...
+      "logoutUrl": "${openmrsBase}/authenticationui/login/login.page"
```
mseaton, line 66: Not really sure why this would be the logoutUrl?  Why wouldn't this be something like `/appui/header/logout.action`?

mseaton (commented):
Few comments @mogoodrich .  Happy to discuss.  Feel free to consider and revise as needed and/or merge otherwise.

configuration/frontend/base-config.json
mogoodrich: Looking at the code, it seems like "logoutUrl" was more "the thing to link to after the OpenMRS logout is complete", which since we don't have any 2 factor place or elsewhere to log out, was just the login page.  However.  it is confusing, and there's no problem with logging out twice, I guess, I will switch to that if it works.

cioan (approved):
besides Mike's comments this looks good to me. Thanks @mogoodrich !

PR #437, opened 2023-09-25
UHM-7454:Disable Patient Lists option in Left-Hand Navigation

mogoodrich:


configuration/frontend/base-config.json
```diff
@@ -52,6 +52,16 @@
   "@openmrs/esm-form-builder-app": {
     "showSchemaSaveWarning": false
   },
+  "@openmrs/esm-home-app": {
+    "extensionSlots": {
+      "homepage-dashboard-slot": {
+        "remove": [
+          "patient-lists-dashboard-link",
+          "home-widget-db-link"
```
mseaton, line 60: Seems fine, but I'm not sold on this approach for O3 in general. I'd favor being declarative here about what to include, rather than say "take whatever is there by default, and remove these 2 things", as you are still left wondering what exactly is being included (which could even change in never versions of the esm...)

PR #441, opened 2023-10-04
HAI-884: Add providers (psychologist and doctor)

louidorjp:
Please review this PR


lnball:
This code looks good, but what's the difference between "provider", "psychologist", and "physician"?  Do you need 3?

image

lnball (approved):
LGTM

PR #442, opened 2023-10-10
change checkboxes into radio buttons and add new concepts

Dutervil:
@mogoodrich @dmdesimone @mseaton  have a look in this please

### DIFF
--- a/configuration/pih/htmlforms/mentalHealth.xml
+++ b/configuration/pih/htmlforms/mentalHealth.xml
@@ -379,34 +379,40 @@
         <enrollInProgram programId="Mental Health" locationTag="Program Location" />
 
         <div class="section-container">
-            <div class="two-columns">
+            
+         <div class="two-columns" id="radio_button_container">
                 <p>
                     <label>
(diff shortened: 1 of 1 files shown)


mogoodrich:
probably the key person to look at this is @lnball ... since this form is in the "PIH EMR" repo, not sure if this is used outside of Haiti?

Dutervil:
@lnball what i do sometimes when the form that  i am  modifying is the in the (openmrs-config-zl), i copy it from where it is and paste to zl I can go fast in terms of running the app. When everything is ok, i go back to the appropriate file [pihemr] to make my change. This was the case for this ticket. So i copied mentalHealth.xml  from [PIH/openmrs-config-pihemr] to zl , so when my code words, i copy may changes and delete the copied file and go back to pihemr and do my work. This is to say the change in made in  **openmrs-config-pihemr file**

Dutervil:
what i do sometimes when the form that i am modifying is **### not**  in the (openmrs-config-zl),

mogoodrich:
Thanks @Dutervil  and @lnball !

PR #449, opened 2023-10-26
Uhm 7482 consolidate bkup restore config

dmdesimone:


mseaton (approved):
This looks fine @dmdesimone although if this file is used in the actual production instance (eg. in an export) then the backup_datetime and restore_datetime will always be null in this case.  Those are only populated during the MySQL restoration process into a separate instance for the warehouse extractions.

PR #450, opened 2023-10-30
UHM-7517, add Mother to Child relationship

cioan:


constants.yml
```diff
@@ -379,6 +379,8 @@ relationshipType:
     uuid: "9a4b3eea-8a9f-11e8-9a94-a6cf71072f73"
   clinicianToPatient:
     uuid: "2fa6fcea-aa58-11e8-98d0-529269fb1459"
+  motherToPatient:
```
mseaton, line 382: It's not really "motherToPatient", it's "motherToChild".  The mother may or may not be a patient, and the child may or may not be a patient.

mseaton (commented):
We should change the name of the constant, otherwise it looks fine (though we tend to have our uuids in lower case).

mseaton (approved):
Looks good, though you'll have to change the uuid to lower case in the other PR too

PR #461, opened 2023-11-15
UHM-7571: Fix OCL load at startup variable

mogoodrich:


configuration/pih/liquibase/liquibase.xml
```diff
@@ -826,4 +826,11 @@
         </sql>
     </changeSet>
 
+    <changeSet id="20231115-update-openconeptlab-gp-after-tomcat-update" author="mgoodrich">
+        <comment>UHM-7571: Fix OCL load at startup variable</comment>
+        <sql>
+            update global_property set property_value = '/home/tomcat/.OpenMRS/ocl/configuration/loadAtStartup' where property = 'openconceptlab.oclLoadAtStartupPath' and property_value LIKE '/home/tomcat7/.OpenMRS/%';
+        </sql>
```
mseaton, line 833: So, I think the better approach here is just to delete the GP altogether.  The code in question in the ocl module will create the GP if it does not exist, and will assign it a value based on the location of the current application data directory.  So a solution that will work for all of our systems, current and future, windows or linux, is one that we just delete the existing GP and always let the OCL module recreate it from scratch at startup using whatever application data directory is currently in place.  The solution above here will only work for a subset of our current implementations and will not necessarily work in the future.

So I'd just change the above to `delete from global_property where property = 'openconceptlab.oclLoadAtStartupPath';`

mseaton (changes requested):
See comments

configuration/pih/liquibase/liquibase.xml
mogoodrich: Makes sense, thanks @mseaton

mogoodrich: Updated, but one thing I discovered... this change set runs after the OCL module startup, so it takes two restarts before the value is set properly.  This may be okay since since we don't really use/case about this GP anyway? Thoughts @mseaton ?

mseaton: Oh, I was wondering that.  Thought maybe all liquibase happened first, then context refreshes/startups happened later.  I think it's fine.  I'm surprised just setting the value to null worked.  I would have though you'd need to delete the GP entirely given the way the code is implemented in the ocl module

mogoodrich: fair enough about switching to just deleting, I made this change, merging, thanks @mseaton

PR #464, opened 2023-11-27
SL-446: Add clinical programs for SL (October 2023) and kgh-test

mogoodrich:


configuration/programs/programs.csv
```diff
@@ -1,17 +1,24 @@
 Uuid,Name,Description,Void/Retire,Program concept,Outcomes concept
```
mogoodrich, line 1: I went ahead and added all these to the main config-pihemr, though it currently it looks like we are primarily adding ZL program here, and Liberia and Mexico programs in their respective configs.  I went back and forth, but seemed like it was better to add programs here (and potentially move the Liberia and Mexico programs here in the future).  The deciding factor to me was that we are the configuring these programs in the CALF, which is shared by all implementations... ie, if I have a component called "familyPlanningProgram", I would assume I could turn that component on just by adding that component to my config... not do that and *also* migrate that program from a site-specific csv to this main csv.  And all the programs generally feel like they *could* be generic, with the exception of the AYFS program.

PR #465, opened 2023-12-01
fixing mental healh page css layout and fixing ltfu translation

Dutervil:
@dmdesimone @lnball fixing mental healh page css layout and fixing ltfu translation


lnball:
@Dutervil Did you fix this on the mental health form?  The label should be above the text area.
image

dmdesimone (commented):
Looks good to me!

PR #468, opened 2023-12-05
SL-462: New National ID field on Registration

mogoodrich:


mogoodrich (commented):
Tangential, I just moved this into messages-sl.properties while adding the new message code for the National ID

mseaton (approved):
LGTM

PR #473, opened 2023-12-15
UHM-7612: add config to disable all-day appointments in O3

chibongho:


mogoodrich (approved):
I don't know if it's just the display, but formatting looks a bit off, otherwise LGTM!

chibongho:
Right... thanks for catching

PR #487, opened 2024-02-02
fixing answer codes translation on Thought content and thought process

Dutervil:
@lnball fixing answer codes translation on Thought content and thought process. Please have a look


lnball (changes requested):
I commented on the ticket (HAI-902).  I would recommend not adding these to the message properties.  Better to add to the concept names.

But if your form uses this label, you will still need it:
- pihcore.mentalhealth.thoughtProcess=Processus de pensée

lnball (commented):
@Dutervil You still have the translations in the message property files?  But you are not using them?

PR #491, opened 2024-02-29
SL-547 - Fix css for datetime of death on htmlforms

mseaton:


mogoodrich (approved):
LGTM, thanks @mseaton

PR #498, opened 2024-03-07
SL-553: add location and privilege for MCOE triage form and visit action

cioan:
Added the privileges and location tags needed to display the new MCOE triage form.

### DIFF
--- a/configuration/locationtags/locationtags.csv
+++ b/configuration/locationtags/locationtags.csv
@@ -16,6 +16,7 @@ d9865139-dfb4-11e4-bccc-56847afe9799,,Vitals Location,Signifies a Location where
 5e824f3a-dfb8-11e4-bccc-56847afe9799,,Inpatients App Location,Signifies a Location where the inpatients app should be available
 24ffeea6-dfbb-11e4-bccc-56847afe9799,,Check-In Location,Signifies a Location where the check-in app and form should be available
 8cff42e5-f03b-461d-b693-65bdf10c4bba,,Check-In Maternal Location,Location where the maternal check-in app and form is available
+a61fa90a-d71d-4b44-b148-affac16fdec3,,MCOE Triage Location,Location where the MCOE triage form is available
 8fa112e0-e506-11e4-b571-0800200c9a66,,Registration Location,Signifies a Location where the registration app should be available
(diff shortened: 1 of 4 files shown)


configuration/locationtags/locationtags.csv
```diff
@@ -16,6 +16,7 @@ d9865139-dfb4-11e4-bccc-56847afe9799,,Vitals Location,Signifies a Location where
 5e824f3a-dfb8-11e4-bccc-56847afe9799,,Inpatients App Location,Signifies a Location where the inpatients app should be available
 24ffeea6-dfbb-11e4-bccc-56847afe9799,,Check-In Location,Signifies a Location where the check-in app and form should be available
 8cff42e5-f03b-461d-b693-65bdf10c4bba,,Check-In Maternal Location,Location where the maternal check-in app and form is available
+a61fa90a-d71d-4b44-b148-affac16fdec3,,MCOE Triage Location,Location where the MCOE triage form is available
```
lnball, line 19: I thought the MCOE Triage Location tag would be added to the sl config, not here.  It's already a location in the sl configuration.

lnball (commented):
See comment @cioan

configuration/locationtags/locationtags.csv
cioan: That's what I thought too @lnball . But there is no locationtag config in the pishsl, and whoever added the Check-In Maternal Location tag, has also added that tag in this same file. Please see the one line above. So, I just wanted to follow the same pattern. Thanks!

mogoodrich: "MCOE" is such a specific term to that implementation, I do agree that it would make sense to create a new location tags file for SL, and add it there instead... *but* an alternatively could be to just rename this "MCH Triage Location" or "Maternal Triage Location" (to be consistent with "Check-In Maternal".

mogoodrich (approved):
See my thoughts on the location/naming convention @cioan thanks!

configuration/locationtags/locationtags.csv
cioan: Thanks @lnball and @mogoodrich ! I have renamed to MCH Triage Location

PR #500, opened 2024-03-19
Uhm 7816   user extracts running slow

dmdesimone:


mseaton (changes requested):
@dmdesimone I believe we decided not to do this in favor of the indexes that you tested.  So I think this can be closed out without merging?

PR #502, opened 2024-03-22
UHM-7577: View Patient Appointments Via O2 Patient Char

mogoodrich:


configuration/messageproperties/messages_en.properties
```diff
@@ -1761,6 +1761,14 @@ pihcore.ncd.other=Other NCDs
 pihcore.hiv.legacyDataNotMigrated=WARNING: This data comes from an old form in the old HIV EMR. Some data which may have been filled in on the old form might not appear here.
 pihcore.hiv.viewPatientInHivEmr=Open patient record in HIV EMR
 
+pihcore.appointments.status.Requested=Requested
+pihcore.appointments.status.WaitList=Wait List
+pihcore.appointments.status.Scheduled=Scheduled
+pihcore.appointments.status.CheckedIn=Checked-In
+pihcore.appointments.status.Completed=Completed
+pihcore.appointments.status.Cancelled=Cancelled
+pihcore.appointments.status.Missed=Missed
```
mogoodrich, line 1770: Now that we are no longer using Transifex here, I forget what our pattern is here... do we manually add these keys to all the language files, just with the English key, or keep them empty?

mseaton: I don't recall.  I would just leave them empty for now, and we have that utility script that I wrote a while back that will help us to identify missing translations if we need it.

mseaton (approved):

PR #503, opened 2024-03-26
UHM-7823 Order obsgroups for newborns

lnball:
Maybe I don't number the controlId?  Unclear how this works...


lnball:
Not used.

PR #511, opened 2024-05-10
SL-613, add strings for the new children page

cioan:


mogoodrich (approved):
LGTM, thanks @cioan !

PR #512, opened 2024-05-13
adding 'additional comment field to mental health section'

Dutervil:
adding 'additional comment field to mental health section
@lnball , Please have a look

### DIFF
--- a/configuration/pih/htmlforms/mentalHealth.xml
+++ b/configuration/pih/htmlforms/mentalHealth.xml
@@ -917,6 +917,31 @@
 
             </div>
         </div>
+
+           <div class="section-container" id="radio_button_container">
+
+            <div class="col">
+                <!-- <p>
+                    <label>
+                        comment
+                    </label>
+                    <obs conceptId="CIEL:163044" type="textarea"/>
+                </p> -->
+                <p class="comment-label"> <uimessage code="pihcore.exam.comment"/> </p> 
+
+                    <p class="comment-field"> 
+
+                        <p> 
+                            <obs conceptId="CIEL:163044" rows="5" cols="25"/> 
+                        </p> 
+
+                    </p> 
+            </div>
+            <div class="col">
+            
+            </div>
+        
+        </div>
     </section>
 
     <section id="mh-suicide" sectionTag="fieldset" headerTag="legend" headerStyle="title"
--- a/configuration/pih/htmlforms/mentalHealth.xml
+++ b/configuration/pih/htmlforms/mentalHealth.xml
@@ -900,12 +900,16 @@
         <div class="section-container" id="radio_button_container">
 
             <div class="col">
-                <p>
-                    <label>
-                        <uimessage code="pihcore.mentalhealth.examination"/>
-                    </label>
-                    <obs conceptId="CIEL:163044" type="textarea"/>
-                </p>
+
+                <p class="comment-label"> <uimessage code="pihcore.exam.comment"/> </p> 
+
+                    <p class="comment-field"> 
+
+                        <p> 
+                            <obs conceptId="CIEL:163044" rows="5" cols="25"/> 
+                        </p> 
+
+                    </p> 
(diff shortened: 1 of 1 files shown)


mogoodrich:
Thanks @Dutervil !  @lnball  in on vacation until next week.  This looks good to me, but I will defer to Ellen as she knows this code better.  That being said, if you need this merged in sooner than next week. let me know and I can approve and we can follow up with Ellen when she is back.

Dutervil:
I think you can merge @mogoodrich , ask you say, Ellen can look at it when She is back

configuration/pih/htmlforms/mentalHealth.xml
```diff
@@ -917,6 +917,31 @@
 
             </div>
         </div>
+
+           <div class="section-container" id="radio_button_container">
```
mogoodrich, line 921: this id seems inaccurate, as this isn't a radio button container, was this copied from somewhere else?

configuration/pih/htmlforms/mentalHealth.xml
```diff
@@ -917,6 +917,31 @@
...
+
+            <div class="col">
+                <!-- <p>
+                    <label>
+                        comment
+                    </label>
+                    <obs conceptId="CIEL:163044" type="textarea"/>
+                </p> -->
```
mogoodrich, line 929: unless we really need to keep it for some reason, let's just remove this instead of commenting it out?

mogoodrich (changes requested):
Thanks @Dutervil ... I am happy to merge and have Ellen review when she gets back, but I just found two small things, can you take a look? Thanks!

configuration/pih/htmlforms/mentalHealth.xml
Dutervil: I have created this id 'radio_button_container' because i wanted to fix the display of the field

configuration/pih/htmlforms/mentalHealth.xml
Dutervil: what you ask me to remove is what we neek i think @mogoodrich

configuration/pih/htmlforms/mentalHealth.xml
mogoodrich: Ah, okay, I see why I was confused.  Element ids should always be unque, but looks like in the mental health form there are many elements with the id "radio_button_container".  Could we change this to a class, ie


        .radio_button_container{
        display:flex;
        gap:50px;
        justify-content:space-between;
        }

... and then change all the id="radio_button_container" to class="radio_button container"?

configuration/pih/htmlforms/mentalHealth.xml
```diff
@@ -917,6 +917,31 @@
...
+                <p class="comment-label"> <uimessage code="pihcore.exam.comment"/> </p> 
+
+                    <p class="comment-field"> 
+
+                        <p> 
+                            <obs conceptId="CIEL:163044" rows="5" cols="25"/>
```
lnball, line 935: @Dutervil You are already using this concept (CIEL:163044) on line 907 of this file.  You cannot use the same concept twice unless it's grouped in an obsgroup/convenience set (ConvSet).

lnball (changes requested):
@mogoodrich made good suggestions about using class instead of id.  

@Dutervil You added a text area with the same concept that you used previously.  That's not allowed.

configuration/pih/htmlforms/mentalHealth.xml
Dutervil: @lnball What is the right concept to use, please?

Dutervil: @lnball I was not sure about the conpcet , i did it while you were in holiday, Please let me know the right ticket to use pleae

lnball (approved):
Looks good.  Thanks.

PR #519, opened 2024-07-17
LIB-335: Rows with blank EMR ID (and everything else except two date)

mogoodrich:


configuration/pih/liquibase/liquibase.xml
```diff
@@ -952,5 +952,15 @@
         </sql>
     </changeSet>
 
+    <changeSet id="20240717-void-programs-associated-with-voided-patients" author="mgoodrich">
+        <comment>LIB-335: Rows with blank EMR ID (and everything else except two dates)</comment>
+        <sql>
```
mseaton, line 957: Would be more accurate to set the date_voided of the patient_program from the date_voided of the patient, and then set the date_voided of the patient_state from the date_voided of the patient_program

mseaton (approved):
Looks fine, though one comment about using now() vs. the indented date_voided from the patient.

configuration/pih/liquibase/liquibase.xml
mogoodrich: I see your point, but, especially since these are audit fields, I prefer to set this to the date the void actually happened, not when it was supposed to happen. This also makes it coincide better with the void_reason ("LIB-335", not "patient merged").

PR #521, opened 2024-08-01
UHM-8098: update multiple edd elements on the same form

cioan:
@mseaton  and @lnball, if you insist on using multiple elements with the same ID, I found another jquery syntax that allows us to select multiple elements whose IDs start with the same string:
**jq("div[id^=calculated-edd-and-gestational]").show();**

Screenshot 2024-08-01 at 3 45 11 PM

### DIFF
--- a/configuration/pih/htmlforms/covid19Intake.xml
+++ b/configuration/pih/htmlforms/covid19Intake.xml
@@ -349,7 +349,7 @@
                             <span id="calculated-edd-label">
                                 <uimessage code="pihcore.calculatedEstimatedDeliveryDate"/>:&#160;
                             </span><br/>
-                            <span id='calculated-edd' class="calculated-edd value"></span>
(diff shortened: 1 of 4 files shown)


configuration/pih/htmlforms/covid19Intake.xml
```diff
@@ -349,7 +349,7 @@
                             <span id="calculated-edd-label">
                                 <uimessage code="pihcore.calculatedEstimatedDeliveryDate"/>:&#160;
                             </span><br/>
-                            <span id='calculated-edd' class="calculated-edd value"></span>
+                            <span id='calculated-edd-value' class="calculated-edd value"></span>
```
mseaton, line 352: I don't insist on using ids @cioan .  If you are going to write code that simulates using classes, you might as well use them.

lnball (approved):
Thanks @cioan .

Can you explain tomorrow how I should use this?  Not sure why you changed these 3 htmlforms.

PR #530, opened 2024-09-20
UHM-8185 Updated form_id for ZL Outpatient Intake (Peds) form in enco…

lnball:
…unter table

Please check the liquibase changeset.

### DIFF
--- a/configuration/pih/liquibase/liquibase.xml
+++ b/configuration/pih/liquibase/liquibase.xml
@@ -972,4 +972,25 @@
         </sql>
     </changeSet>
 
+    <changeSet id="20240920-migrate-outpatient-peds-form-id" author="ball">
+        <preConditions onFail="MARK_RAN">
+            <sqlCheck expectedResult="1">
+                select count(*) from address_hierarchy_entry where parent_id is NULL and name like 'Haiti';
+            </sqlCheck>
+        </preConditions>
+        <comment>
+            If the count is 1, this is a Haiti server
+        </comment>
+        <sql>
+            -- Update to Outpatient Initial (Peds) form_id when the encounter_type is Outpatient Initial
+            update encounter
+            set form_id =
+                    (select form_id from form where uuid = '58f731af-8f8e-4863-b721-70bca76699c4')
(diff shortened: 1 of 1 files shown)


configuration/pih/liquibase/liquibase.xml
```diff
@@ -972,4 +972,25 @@
         </sql>
     </changeSet>
 
+    <changeSet id="20240920-migrate-outpatient-peds-form-id" author="ball">
+        <preConditions onFail="MARK_RAN">
+            <sqlCheck expectedResult="1">
+                select count(*) from address_hierarchy_entry where parent_id is NULL and name like 'Haiti';
```
mogoodrich, line 978: We also need a precondition to check  that select form_id from form where uuid = ''58f731af-8f8e-4863-b721-70bca76699c4' is not null.

mogoodrich (changes requested):
Generally looks good, though you will need to add a precondition or this will just end up setting the form_id for all these forms to null

configuration/pih/liquibase/liquibase.xml
```diff
@@ -974,15 +974,16 @@
 
     <changeSet id="20240920-migrate-outpatient-peds-form-id" author="ball">
         <preConditions onFail="MARK_RAN">
```
mogoodrich, line 976: See my comments on the ticket, but I think this should be "onFail="CONTINUE""

configuration/pih/liquibase/liquibase.xml
```diff
@@ -974,15 +974,16 @@
...
-            <sqlCheck expectedResult="1">
-                select count(*) from address_hierarchy_entry where parent_id is NULL and name like 'Haiti';
+            <sqlCheck expectedResult="0">
```
mogoodrich, line 977: ... and the expected result should be "1"

mogoodrich (changes requested):
See my comments on the ticket for my logic, let me know if that makes sense, thanks @lnball

mogoodrich (approved):
LGTM, thanks @lnball !

PR #533, opened 2024-10-16
SL-771: Default the pregnancy and infant patient dashboard for certai…

mogoodrich:
…n user locations

### DIFF
--- a/configuration/locationtags/locationtags.csv
+++ b/configuration/locationtags/locationtags.csv
@@ -42,5 +42,5 @@ ae29295b-2795-11ec-9efc-0242ac110002,,Echocardiogram Location,
 278ba6a4-7570-11ec-be0e-0242ac110002,,Health Facility Location,"Signifies Locations that represent Health Facilities"
 cc104bff-6eaf-452e-a192-400d189cd82a,,Labs Component Location,"Signifies Locations where the Labs component should be enabled"
 26eac54e-7e6f-4d59-9afd-6cc8362be65a,,NICU Location,Neonatal intensive care unit,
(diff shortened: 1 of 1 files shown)


configuration/locationtags/locationtags.csv
```diff
@@ -42,3 +42,5 @@ ae29295b-2795-11ec-9efc-0242ac110002,,Echocardiogram Location,
 278ba6a4-7570-11ec-be0e-0242ac110002,,Health Facility Location,"Signifies Locations that represent Health Facilities"
 cc104bff-6eaf-452e-a192-400d189cd82a,,Labs Component Location,"Signifies Locations where the Labs component should be enabled"
 26eac54e-7e6f-4d59-9afd-6cc8362be65a,,NICU Location,Neonatal intensive care unit,
+d4e5f6a7-b8c9-0d1e-2f3a-4b5c6d7e8f9a,,Pregnancy Program Location,Locations where the Pregnancy Program is provided
+e5f6a7b8-c9d0-1e2f-3a4b-5c6d7e8f9a0b,,Infant Program Location,Locations where the Infant Program is provided
```
mogoodrich, line 46: We could potentially call these "Pregnancy Program *Dashboard* Location"  and "Infant Location *Dashboard* Location" if we wanted to make them more specific. It makes it clearer what the tag used for, but also means we'd need a new tag (or this tag would become misnamed) if we wanted to start using it for something else.

mogoodrich: The locations that were requested to be tagged with these tags were also a broader set that the ones with the Maternal and Child Health tag.  I might be leaning now towards adding "Dashboard" to the names of these tags to make it 100% clear.

mseaton (approved):
LGTM, defer to you on the naming

PR #535, opened 2024-10-18
SL-770: Pregnancy Treatment state outcomes

mogoodrich:


configuration/programworkflowstates/pregnancy.csv
```diff
@@ -0,0 +1,2 @@
+${programWorkflow.pregnancyProgramTypeOfTreatment.state.prenatalIndividual.uuid},,${programWorkflow.pregnancyProgramTypeOfTreatment.uuid},PIH:11700,true,
+${programWorkflow.pregnancyProgramTypeOfTreatment.state.postpartum.uuid},,${programWorkflow.pregnancyProgramTypeOfTreatment.uuid},PIH:13137,true,
```
mogoodrich, line 2: It looks like we are attempting to override these default states in the pregnancy program in the SL config, so I pulled these out into a separate file called "pregnancy.csv" so we could override them with a "pregnancy.csv" in config-pihsl (see related PR there)

mogoodrich (commented):
FYI @lnball  going to merge this in, but we can discuss in detail when you are back.

PR #536, opened 2024-10-21
HAI-1004:  make test date field non-mandatory

louidorjp:
Please review this PR

### DIFF
--- a/configuration/pih/htmlforms/labResults.xml
+++ b/configuration/pih/htmlforms/labResults.xml
@@ -227,8 +227,8 @@
                     <uimessage code="pihcore.lab.test_date"/>
                 </label>
                 <div class="small">
-                    <field class="required">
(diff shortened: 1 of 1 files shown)


mogoodrich (approved):
This generally makes sense to me @louidorjp , but a question for @lnball  and @dmdesimone  do you know if we use this form outside of Haiti and, if so, for those other sites to we need to maintain a separate version of this form with the date of test results mandatory?

dmdesimone:
> This generally makes sense to me @louidorjp , but a question for @lnball and @dmdesimone do you know if we use this form outside of Haiti and, if so, for those other sites to we need to maintain a separate version of this form with the date of test results mandatory?

Actually, we just discussed this in a call.  We agreed **NOT** to make this change since it will cause more problems in reporting if results are added without a test result date.  Instead we will look into adding Test Reason to the Lab Order page so VL reporting with test reason can be done before the results have been added.  @louidorjp @mogoodrich @lnball

PR #545, opened 2024-11-21
UHM-8236: Review Program Workflow Iniz Deployment

mogoodrich:
see: https://pihemr.atlassian.net/browse/UHM-8236


mseaton (approved):

constants.yml
```diff
@@ -434,6 +435,18 @@ programWorkflow:
   infantProgramEidTreatmentStatus:
     # this is referred to by UUID because both eid and infant programs have the workflow with the same concepts
     uuid: "29aba153-4f1e-4c28-bd36-0332b30fe4b5"
+  mchProgramTypeOfTreatment:
```
lnball, line 438: You want this in pihemr?  That's ok with me.

mogoodrich: Yeah, I think?  Just left a comment on the other ticket... we can try this and move if it doesn't work out.

PR #546, opened 2024-11-22
UHM-8282: Dispensing Form "Location of Prescription" Shows all Locations

mogoodrich:


lnball:
Sounds better @mogoodrich 

Could we  use “Dispensing Location” instead of “Login Location”?  Or will “Dispensing Locations” show non-HUM locations?

mogoodrich:
Thanks @lnball see my new comment on the ticket re: your question.

PR #549, opened 2024-12-05
I finish the form

Dutervil:
Please @lnball , have a look at this


configuration/pih/scripts/global/mch.js
```diff
@@ -27,11 +27,11 @@ function setUpEdd(currentEncounterDate, msgWeeks) {
       let yearMonthDay = lastPeriodDateValue.split('-');
       if (yearMonthDay.length == 3) {
         lastPeriodDate.setFullYear(yearMonthDay[0]);
-        lastPeriodDate.setMonth(+yearMonthDay[1] -1); // the month starts from 0 for January
+        lastPeriodDate.setMonth(+yearMonthDay[1] - 1); // the month starts from 0 for January
```
lnball, line 30: Does this break this?  Do you need this space?

configuration/pih/htmlforms/mentalHealth.xml
lnball, line 1: I suggest creating this file in the zl repo.  Do not change the one in openmrs-config-pihemr.  Same issue with the vitals.  It's possible that the vtials form is used by other countries.  (I think that Sierra Leone and Liberia have their own mental health form).

lnball (changes requested):
@Dutervil See my comments.

configuration/pih/scripts/global/mch.js
Dutervil: I did not touch this code

lnball: It's showing the line changed by you.  Just remove the extra space.

image

Dutervil: Sure . I will

PR #559, opened 2025-01-03
SL-816 Duplicate registrations caused by system error

dmdesimone:


configuration/pih/liquibase/sql/void_duplicate_registrations.sql
```diff
@@ -0,0 +1,158 @@
...
+	order by preferred desc, date_created desc limit 1)
+set t.address_level_1 = a.country,
+	t.address_level_2 = a.state_province,
+	t.address_level_3 = a.city_village,
+	t.address_level_4 = a.county_district ,
+	t.address_level_5 = a.address1;
+
+-- person attributes
+select person_attribute_type_id into @telephone from person_attribute_type where name = 'Telephone Number' ;
+select person_attribute_type_id into @motherName from person_attribute_type where name = 'First Name of Mother' ;
```
mogoodrich, line 64: Not a big deal, but do we connect these attributes in all our countries?  I guess if not just needs to be null-safe.

configuration/pih/liquibase/sql/void_duplicate_registrations.sql
```diff
@@ -0,0 +1,158 @@
...
+and d.telephone_number <=> p.telephone_number
+and d.address_level_1 <=> p.address_level_1
+and d.address_level_2 <=> p.address_level_2
+and d.address_level_3 <=> p.address_level_3
+and d.address_level_4 <=> p.address_level_4
+and d.address_level_5 <=> p.address_level_5
+and d.reg_location <=> p.reg_location
+and d.registration_date <=> p.registration_date
+and date(d.registration_entry_date) <=> date(p.registration_entry_date)
+and d.user_entered <=> p.user_entered;
```
mogoodrich, line 133: I'm probably missing this, but why do we have to do a comparison on all the fields again?  Doesn't the group part handle that?

configuration/pih/liquibase/sql/void_duplicate_registrations.sql
```diff
@@ -0,0 +1,158 @@
...
+-- do not have any encounters but where the other patients in that group have encounters (excluding registration)   
+DROP TEMPORARY TABLE IF EXISTS temp_dups_to_void;
+CREATE TEMPORARY TABLE  temp_dups_to_void
+select patient_id from temp_potential_dups d 
+where not exists
+ (select 1 from encounter e where  e.patient_id = d.patient_id and e.encounter_type <> @regEncounterType)
+and exists 
+ (select 1 from encounter e2 where e2.patient_id in
+ 	(select patient_id from temp_potential_dups_patient_ids d2 where d2.group_id = d.group_id)
+ 	and e2.encounter_type <> @regEncounterType
```
mogoodrich, line 149: I'm probably missing this, but do you need  to confirm that e2.patient is not equalt to e.patient?  Or I guess it's the fact that the two clauses can never be both true for the same patient?

mogoodrich (approved):
Had some questions, but generally LGTM! 

Also, feel free to assign this back to me and/or do a other ticket about looking into the root cause of the problem again.

configuration/pih/liquibase/sql/void_duplicate_registrations.sql
dmdesimone: Right - I made sure that things were null-safe.

configuration/pih/liquibase/sql/void_duplicate_registrations.sql
dmdesimone: This is the part that, once you identify all the duplicate rows without the patient_id, it joins these rows back to the original dataset (that HAS patient_id) so in order to join to the correct row, you need to make sure every field matches.  

I know there more straightforward ways to do it, but after lots of trial and error, by far the fastest way to do it is to group by all the fields you want to compare first.

configuration/pih/liquibase/sql/void_duplicate_registrations.sql
dmdesimone: right... I didn't think to do it because it would need to be a different patient for the two clauses to be true.  Shouldn't need it.

PR #561, opened 2025-01-10
SL-838 - Forms and Locations for MCOE

mseaton:


mogoodrich (approved):
LGTM

So we decided to stick with thematic locations rather than locations that are meant to map directly to a certain encounter type?  (fine with me, but if we do mean for them to map to an encounter type I think I would prefer to include "Encounter" in the tag name)

PR #569, opened 2025-01-28
SL-721 - Configure cause of death from sets in SL

mseaton:


mogoodrich (approved):
This is the new cause of death set we want to sue, replacing the cause of death question?

Looks good to me, but maybe we want to merge the Core Apps cause of death change first, confirm there are no regressions, and then update this and test again?

PR #583, opened 2025-02-27
UHM-8402: fix the calculation of the gestational age

cioan:
@mogoodrich , here is the fix for this function:
Screenshot 2025-02-27 at 10 53 53 AM


mogoodrich (approved):
LGTM!  (defer to others on what the exact definition of gestational age is)

PR #585, opened 2025-03-11
SL-859: Export an all patients list nightly and save offline

mogoodrich:


mseaton (commented):
Preumably the filenameBase is still something the renderer would use to influence the name of the file to be download, and we should keep this?

mogoodrich:
> Preumably the filenameBase is still something the renderer would use to influence the name of the file to be download, and we should keep this?

Fair.  I'm not sure if it actually does, but I'll keep these for now

PR #586, opened 2025-03-13
UHM-8404 - Set date of last menstruation to required but only if visible

mseaton:


configuration/pih/htmlforms/section-obgyn-initial.xml
```diff
@@ -544,13 +544,23 @@
...
+					}
+				});
+				return ( numErrors == 0 );
+			});
+
+			jq('#intake-or-followup').change(function() {
+				setRequiredIfVisible('knowing_period_question');
+			});
+			jq('#knowing_period_question').change(function() {
+				setRequiredIfVisible('knowing_period_question');
```
mogoodrich, line 560: Why do we run this method when these two things elements change?

configuration/pih/htmlforms/section-obgyn-initial.xml
```diff
@@ -544,13 +544,23 @@
...
 			});
-			return ( numErrors == 0 );
+			beforeSubmit.push(function() {
+				return setRequiredIfVisible('knowing_period_question');
```
mogoodrich, line 563: This is more like "validateAsRequiredIfVisible", correct? (admitedly a mouthful of name)

mogoodrich (approved):
I had some questions, but I'm probably just not understanding this fully.  :)

configuration/pih/htmlforms/section-obgyn-initial.xml
mseaton: Sure, I can rename this.

configuration/pih/htmlforms/section-obgyn-initial.xml
mogoodrich: Ah, I think i get it, in this case the method is just being used to display the "Required" text if needed? (hence setRequiredIfNeeded method name?)

mseaton: Yes, that is the main intent.  But it also returns a boolean indicating whether it is currently required and not completed, which allows the same method to be used in the `beforeSubmit` validations that prevent form submission

PR #587, opened 2025-03-14
UHM-8445 - Make perineal laceration section conditional

mseaton:


mogoodrich (approved):
Ah, I understand now @mseaton after seeing the code. This makes sense, and, yes, this is the correct way to fix this, I would say that the old code was "working as intended" from a HFE perspective.  :)

lnball:
It won't work inside a repeat tag but works with this?

PR #588, opened 2025-03-25
UHM-8473 Liquibase changeset for duplicate trauma data on mental heal…

lnball:
…th form

### DIFF
--- a/configuration/pih/liquibase/liquibase.xml
+++ b/configuration/pih/liquibase/liquibase.xml
@@ -1119,11 +1119,23 @@
         </sql>
     </changeSet>
 
-
-    <changeSet id="20250325-void-duplicate-trauma-obs" author="ball">
+    <changeSet id="20250326-void-duplicate-trauma-obs" author="ball">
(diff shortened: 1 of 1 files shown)


configuration/pih/liquibase/liquibase.xml
```diff
@@ -1118,5 +1118,45 @@
...
         </sql>
     </changeSet>
+
+
+    <changeSet id="20250325-void-duplicate-trauma-obs" author="ball">
+        <preConditions onFail="MARK_RAN">
+            <sqlCheck expectedResult="1">
+                select count(*) from address_hierarchy_entry where parent_id is NULL and name like 'Haiti';
+            </sqlCheck>
+        </preConditions>
```
dmdesimone, line 1128: wouldn't this run EVERY time for Haiti?  Should we put a count of the duplicate obs here in the precondition so it only runs when it needs to?

dmdesimone (approved):
Looks good with my one comment

PR #594, opened 2025-04-09
UHM-8478 Added NCD forms for REDCap

lnball:
Added Version 0.5 of the NCD forms (intake and followup).  I was able to test the intake form and it showed up fine.  We can make adjustments to the sections but I don't think you need all the same sections that are in the NCD 1.0 (ie. exams and sickle cell).

I will make one more tweak of the encounterTypeConfig now for the followup form.  Give me 5 minutes.

We will still need to update the ncd V 0.5 to match REDCap.


cioan (approved):
Thanks @lnball !

mogoodrich (approved):

PR #608, opened 2025-05-21
SL-917: add require expression functions for appframework configuration

cioan:
@mseaton , please let me know if there is a better place to store those java script functions. Thanks!


configuration/appframework/scripts/isPatientEnrolledInProgramOfUuid.js
```diff
@@ -0,0 +1,6 @@
+function isPatientEnrolledInProgramOfUuid(programUuid) {
```
mseaton, line 1: Let's name this more concisely - we don't need `InProgramOfUuid` in the file name or method name

configuration/appframework/scripts/patientHasEncounterOfTypeDuringProgramEnrollment.js
```diff
@@ -0,0 +1,8 @@
+function patientHasEncounterOfTypeDuringProgramEnrollment(encounterTypeUuid, programUuid) {
+
+  var isPatientEnrolled = isPatientEnrolledInProgramOfUuid(programUuid);
+  if ( !isPatientEnrolled ) {
+    return false;
+  }
+  return isPatientEnrolledInProgramOfUuid(programUuid) && encounterOfTypeAfterDate(encounters, encounterTypeUuid, getDateEnrolledByProgramUuid(activePrograms, programUuid));
```
mseaton, line 7: You don't need to evaluate `isPatientEnrolledInProgamOfUuid` again here, you already did it above.  Can't we just have this final statement?

cioan:
Thanks @mseaton ! I have updated the PR based on your review.

configuration/appframework/scripts/getDateEnrolledByProgramUuid.js
```diff
@@ -0,0 +1,12 @@
+function getDateEnrolledByProgramUuid(patientPrograms, programUuid) {
+  if (!patientPrograms) {
+    return false;
+  }
+  var i, len = patientPrograms.length;
+  for (i = 0; i < len; ++i) {
```
mseaton, line 6: See comment above - use `patientPrograms.forEach`, rather than all of these variables.

configuration/appframework/scripts/getDateEnrolledByProgramUuid.js
```diff
@@ -0,0 +1,12 @@
...
+    if (patientPrograms[i]['programUuid'] === programUuid) {
+      return patientPrograms[i]['dateEnrolled'];
+    }
+  }
+  return false;
```
mseaton, line 11: Same as above.  Why are you returning a boolean here?

configuration/appframework/scripts/isPatientEnrolledInProgramOfUuid.js
```diff
@@ -0,0 +1,6 @@
+function isPatientEnrolled(programUuid) {
```
mseaton, line 1: You need to change this file name to match the function name, now that you have changed it.  Also, this is inconsistent with the way you have implemented the other functions.  The way you have implemented the others is to have all of the arguments you need passed into the function, whereas this one assumes some arguments are already defined as global variables.  We should choose one of these approaches and stick to it for consistency.

mseaton:
See my new comments @cioan

configuration/appframework/scripts/encounterOfTypeAfterDate.js
```diff
@@ -0,0 +1,17 @@
...
+  if (!encounters) {
+    return false;
+  }
+
+  if (!afterDate) {
+    return false;
+  }
+
+  var i, len = encounters.length;
+  for (i = 0; i < len; ++i) {
```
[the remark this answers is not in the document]
cioan: @mseaton, sure, but is this javascript syntax supported by the engine? I was trying to avoid any newer methods and syntax.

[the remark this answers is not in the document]
mseaton: @cioan you might be right but might as well check?  If it isn't supported, you could use the `some` function like you do in the `isPatientEnrolled` function, right?

cioan:
Thanks @mseaton ! I think I addressed all your comments above, except .each() which I did not get a chance to test to see if it's supported by the engine.

configuration/appframework/scripts/getDateEnrolledByProgramUuid.js
```diff
@@ -0,0 +1,12 @@
+function getDateEnrolledByProgramUuid(patientPrograms, programUuid) {
+  if (!patientPrograms) {
+    return null;
+  }
+  var i, len = patientPrograms.length;
+  for (i = 0; i < len; ++i) {
+    if (patientPrograms[i]['programUuid'] === programUuid) {
+      return patientPrograms[i]['dateEnrolled'];
```
mseaton, line 8: What if the patient has multiple enrollments for the same program?  What happens then?  Could easily happen in the pregnancy program.  What is the use case for this function and can we make it more explicit?

configuration/appframework/scripts/isPatientEnrolled.js
```diff
@@ -0,0 +1,6 @@
+function isPatientEnrolled(programs, programUuid) {
```
mseaton, line 1: Based on your logic here, I'd name this something like `isEverEnrolled`.  If this is intended to indicate if there is a _current_ / _active_ enrollment, then that is not being checked here.  `isPatientEnrolled` implies that they are currently enrolled.

configuration/appframework/scripts/patientHasEncounterOfTypeDuringProgramEnrollment.js
```diff
@@ -0,0 +1,12 @@
+function patientHasEncounterOfTypeDuringProgramEnrollment(programs, encounterTypeUuid, programUuid) {
+
+  var patientEnrolled = isPatientEnrolled(programs, programUuid);
+  if ( !patientEnrolled ) {
+    return false;
+  }
+  var dateEnrolled = getDateEnrolledByProgramUuid(programs, programUuid);
+  if ( !dateEnrolled ) {
+    return false;
+  }
```
mseaton, line 10: I don't see the point in checking both of these things above.  All you need is the `dateEnrolled`.  No point in checking `isPatientEnrolled` when a null `dateEnrolled` gives you the same information.

cioan:
@mseaton , I found it to be much clearer to just have one function that checks if patient has an encounter of given type during a given program enrollment. I have removed all other script functions. Now, since I have the entire logic in one function, and tested successfully with multiple patient visits and program enrollments, I am happy to break it down in smaller functions if you think that it is still necessary?

configuration/appframework/scripts/patientHasEncounterOfTypeDuringProgramEnrollment.js
```diff
@@ -0,0 +1,29 @@
+function patientHasEncounterOfTypeDuringProgramEnrollment(patientPrograms, programUuid, encounters, encounterTypeUuid, visit) {
+
+  if ( (typeof patientPrograms === 'undefined') || !patientPrograms || (typeof encounters === 'undefined') || !encounters) {
+    return false;
+  }
+
+  var foundProgram = null;
+  for ( var i = 0; i < patientPrograms.length ; ++i ) {
+    if (( patientPrograms[i].programUuid === programUuid ) && visit
```
mseaton, line 9: It seems that `visit` is required for this to work, so shouldn't we move the `&& visit` to line 3 with the other checks?

configuration/appframework/scripts/patientHasEncounterOfTypeDuringProgramEnrollment.js
```diff
@@ -0,0 +1,29 @@
...
+      && ( (new Date(visit.startDatetimeInMilliseconds)) >= ( new Date(patientPrograms[i].dateEnrolled)))
+      && ( patientPrograms[i].dateCompleted == null || (new Date(patientPrograms[i].dateCompleted)) >= (new Date(visit.startDatetimeInMilliseconds))) ) {
```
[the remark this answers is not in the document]
cioan: Thanks @mseaton ! I have updated the PR with your suggestion.

mseaton (approved):

PR #615, opened 2025-05-29
SL-960 remove order basket from ward app

chibongho:


mogoodrich (approved):

chibongho:
Sorry, that didn't work. The entry got override by the SL config. I should have done this there instead. Reverting.

PR #618, opened 2025-06-04
HAI-1046: MCH - OBGYN : Ananlysis section added

louidorjp:
{B218EB50-1E63-419D-B3DF-5AE69E810E49}
Adding the Analysis section to the OBGYN form.


mogoodrich:
fyi @fiona855

PR #619, opened 2025-06-06
UHM-8242 hide O3 left nav in home app and dispensing app

chibongho:
Adding config to hide the left nav in the home app and dispensing app.


mseaton (approved):
This is fine - assuming we sure we want this across the pihemr vs. just in SL

PR #621, opened 2025-06-24
SL-965 add hotjar for SL training server

chibongho:
This is ready for review, but marking this as draft, as IT is in the process of making a external domain for this server, so the hard-coded address `"sle-kgh-inf-omrstest"` will change.


lnball (commented):
LGTM but you said you are awaiting IT for the hostname/domain, @chibongho

PR #630, opened 2025-08-20
HAI-736: Afficher le site auquel appartient un patient lors de la rec…

mogoodrich:
…herche (Show the site a patient belongs to when searching)


mogoodrich:
We will use this to store a patient's Health Center in the Haiti HIV system... this is a person attribute included in the standard OpenMRS setup, I believe, but we will directly add it here since we are going to start to use it.  Adding the constant also allows us to reference it via the constant in PIH Core.

PR #638, opened 2025-09-22
HAI-1062 - Replace lab test observations with orders on the obgyn form

mseaton:


configuration/pih/scripts/visit/encounterTypeConfig.js
```diff
@@ -2011,20 +2011,37 @@ angular.module("encounterTypeConfig", [])
...
+                    showOnVisitList: true,
+                    sections: [
+                        obgynInitial,
+                        ancVaccinations,
+                        primaryCareExam,
+                        obgynDx,
+                        obgynPlan
+                    ]
+                },
+                1.0: {
```
mogoodrich, line 2030: Oh, annoying, I forgot that we *did* end up having to add support to modifying the encounterTypeConfig to get form versions to work properly?  it's possible you only need to do this if you are changing the sections that appear on a form in a new version?

mseaton: Hmm.  Yeah, it's possible that this change wasn't actually needed in the end.

PR #710, opened 2026-04-30
Claude-generated HTML Form documentation

mogoodrich:
I think Dave may have already done something like this, and I haven't really reviewed this at all, but just wanted to push this up as an example of Claude is capable of.  Likely we can get something better with better prompting.

I don't think we should merge this in, but happy to continue to iterate.

### DIFF
--- a/htmlforms-documentation.md
+++ b/htmlforms-documentation.md
@@ -1,0 +1,810 @@
+# PIH EMR HTML Forms Documentation
+
+This document describes all XML forms in `configuration/pih/htmlforms/`, organized by clinical area.
+
+---
+
+## Table of Contents
+
+1. [Core Clinical Forms](#1-core-clinical-forms)
+2. [Maternal & OB-GYN Forms](#2-maternal--ob-gyn-forms)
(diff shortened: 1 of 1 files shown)


lnball:
Cranks out some very useful documentation.  Thanks for your experimentation.

lnball (approved):
There are a couple of things that need to be reviewed but it seems 99% right.

mogoodrich:
@lnball you think enough so that we should tweak and get this merged in?  (fyi @dmdesimone  @mseaton )

lnball:
@mogoodrich Is there a ticket for this?  Can you assign to me?  1 point but enough for me to keep on my radar in the next days.

mogoodrich:
> @mogoodrich Is there a ticket for this? Can you assign to me? 1 point but enough for me to keep on my radar in the next days.

Fair @lnball !  This started out as a 5-minute-at-the-end-of-the-day thing, but we should ticket at this point: 
https://pihemr.atlassian.net/browse/UHM-9281

lnball (approved):
I will merge and edit.  Looks like a nice addition.  Will be interesting to see how much editing is needed.

PR #712, opened 2026-05-04
HAI-1173 : New component nomenclature

Dutervil:


lnball (changes requested):
@Dutervil These are "outpatient" forms and used at Klinik Eksten.  They could be used at inpatient but that was NOT the purpose of these forms.

PR #715, opened 2026-05-12
SL-1333: Populate past Inborn visit attributes

cioan:
This ran successfully on my local kgh-test instance.

### DIFF
--- a/configuration/pih/liquibase/sql/set-inborn-visit-attribute.sql
+++ b/configuration/pih/liquibase/sql/set-inborn-visit-attribute.sql
@@ -1,0 +1,38 @@
+SET @infant_concept_uuid       = '23eeeec5-7f82-4bea-8bdf-f959900882e7';
+SET @birthtime_concept_uuid      = '5599AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA';
+SET @inborn_visit_attribute_type_uuid = '86f716fc-5e26-4eb1-9484-46370cff28f0';
+
+DROP TABLE IF EXISTS tmp_inborn_visit;
+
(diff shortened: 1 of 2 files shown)


configuration/pih/liquibase/sql/set-inborn-visit-attribute.sql
```diff
@@ -0,0 +1,38 @@
...
+JOIN concept c1 ON o1.concept_id = c1.concept_id
+JOIN obs o2    ON o2.obs_group_id = o1.obs_group_id
+JOIN concept c2 ON o2.concept_id = c2.concept_id
+JOIN person p  ON p.uuid = o1.comments
+JOIN visit v   ON v.patient_id   = p.person_id
+              AND v.date_started = o2.value_datetime
+WHERE c1.uuid = @infant_concept_uuid
+  AND o1.comments REGEXP '^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$'
+  AND c2.uuid = @birthtime_concept_uuid
+  AND o1.obs_group_id IS NOT NULL;
```
mogoodrich, line 19: This looks right, but would be good to sanity check against a patient that has given birth more than once to make sure it doesn't get confused.

mogoodrich (approved):
Thanks @cioan !  One comment but generally looks good.

configuration/pih/liquibase/sql/set-inborn-visit-attribute.sql
cioan: Thanks @mogoodrich ! I did test it with twins as well.

PR #717, opened 2026-05-13
UHM-9155 - Initial pass at medication dispense label report

mseaton:


mogoodrich (approved):
So basically we specify a template via HTML and there's a tool that converts it to HTML to PDF?  Cool.

PR #727, opened 2026-05-29
UHM-9325: Fix Civil Status foreign key -- person attribute

mogoodrich:


mseaton (approved):
Sure, though I'd be happier if we didn't store this as a person attribute at all

lnball (approved):
We had discussed in 2012 with Darius.  There were the 2 choices -- person attribute or obs.  We decided this was better attached to the patient and not an encounter/form.  But I know it's collected on the registration form.  I don't think there's an advantage to changing it now @mseaton

PR #731, opened 2026-06-22
UHM-9355: do not erase existing EDD values in the EDIT mode

cioan:
Thanks @mseaton for diagnosing this problem. Here is the fix to the this function. Thanks!

Screenshot 2026-06-22 at 10 38 53 AM

The key idea: encObsEdd is captured once at setup, so it can't tell "load" from "user edit." The new isInitialLoad flag does — guards only apply on the initial pass; any user-driven LMP change runs the full
  recalculation. Gestational age uses its own guard (preserveExistingGestAge) so the two fields stay independent, matching the original code's separate obs checks.
  - When a user changes the LMP on an EDIT form that had a saved EDD, the recalculated EDD will now overwrite the saved one.


mseaton (approved):
Looks ok, will need thorough testing

cioan:
Thanks @mseaton !

PR #732, opened 2026-06-22
UHM-9321 Created starter code for estimated delivery date (and GA)

lnball:
This is a starter.  I think we should have another xml for followup or show-edd-and-ga, but they should share the same css and js files.

I think we still need to fix this code because there are still a few problems.  But hopefully if we pull the code into a few files it will be easier to maintain and modify.

I am also adding PRs for the sl htmlform files.

### DIFF
--- a/configuration/pih/subforms/est-delivery-date-ga-initial.xml
+++ b/configuration/pih/subforms/est-delivery-date-ga-initial.xml
@@ -1,0 +1,60 @@
+<!-- The htmlform code for Initial estimated delivery date and gestational age -->
+<div id="edd-and-ga-widget">
+    <subform path="configuration/pih/subforms/est-delivery-date-ga.js"/>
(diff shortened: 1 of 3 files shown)


cioan:
Thanks @lnball ! You do not need to include the javascript code. I think that more important is that the UI looks consistently in all those forms that have the EDD field. Once we agree on the UI, then it should be straightforward to add the javascript needed to implement the desired behaviour.

lnball:
@cioan I pulled the javascript from the section-anc-intake.xml.

Fiona and I discussed last week and agreed on a consistent layout for intake.  It's also in the description of the ticket:

image

Followup will be slightly different.

Happy to discuss tomorrow or whenever.

mseaton:
One thing I'd say is that once we have these in subforms, including a js subform, we should move _all_ the JS that this uses in there, rather than calling the js in the global/scripts file.  This will allow us to no longer include those JS files in every gsp page in the entire application, and will keep this code where it is used.

PR #734, opened 2026-06-25
SL-1290 remove schedule appointment subform

chibongho:


mogoodrich (approved):

cioan (approved):

cioan:
@chibongho , since you removed the schedule-appointment.xml, the maternalDischarge.xml and maternalDischarge-print.xml forms should also be updated? I assume those are the only forms where this subform was used?

mseaton (approved):

chibongho:
> @chibongho , since you removed the schedule-appointment.xml, the maternalDischarge.xml and maternalDischarge-print.xml forms should also be updated? I assume those are the only forms where this subform was used?

Yes, the corresponding change in those forms are here: https://github.com/PIH/openmrs-config-pihsl/pull/324. I'll merge them all in at the same time after they are approved..

PR #737, opened 2026-07-17
HAI-1126:  Segregate Data Exports by Location

mogoodrich:


configuration/reports/reportdescriptors/dataexports/sql/allpatientswithids.sql
```diff
@@ -30,6 +32,15 @@ select patient_id from patient where voided = 0
...
 
+-- health center (person attribute)
+update temp_patients t
+left outer join person_attribute pa_hc on pa_hc.person_id = t.patient_id
+	and pa_hc.voided = 0
+	and pa_hc.person_attribute_type_id = (select person_attribute_type_id from person_attribute_type where uuid = '8d87236c-c2cc-11de-8d13-0010c6dffd0f')
+left outer join location loc_hc on loc_hc.location_id = pa_hc.value
+set t.health_center_id = pa_hc.value,
+	t.health_center = loc_hc.name;
+
```
mogoodrich, line 43: Add health center person attribute to the report.

configuration/reports/reportdescriptors/dataexports/sql/chiefComplaint.sql
```diff
@@ -24,16 +24,18 @@ insert into temp_cc (
...
   et.name
 from encounter e
 inner join encounter_type et on et.encounter_type_id = e.encounter_type
+inner join visit v on e.visit_id = v.visit_id and v.voided = 0
 where e.encounter_type in (@AdultInitEnc, @AdultFollowEnc, @PedInitEnc, @PedFollowEnc)
  AND date(e.encounter_datetime) >=@startDate
  AND date(e.encounter_datetime) <=@endDate
-and voided = 0
+and e.voided = 0
+AND v.location_id = @location
```
mogoodrich, line 38: Standard pattern for joining on visit, and then limiting by location, repeated often throughout.

configuration/reports/reportdescriptors/dataexports/sql/labResultsExport.sql
```diff
@@ -149,13 +149,17 @@ create index temp_lab_patient_ei on temp_lab_encounter(encounter_id);
...
-	t.specimen_collection_date = e.encounter_datetime, 
+ 	t.encounter_type_id = e.encounter_type,
+	t.specimen_collection_date = e.encounter_datetime,
 	t.date_created = e.date_created,
 	t.creator = e.creator,
-	t.visit_id = e.visit_id;
+	t.visit_id = e.visit_id
+where v.location_id = @location;
+
+DELETE FROM temp_labresults WHERE encounter_id NOT IN (SELECT encounter_id FROM temp_lab_encounter WHERE visit_id IS NOT NULL);
```
mogoodrich, line 162: Ugh, missed this in the first pass but this seems horribly inefficent--looks like the report fetches *all* lab result obs across all patients, and then deletes the ones that are from encounters that aren't associated with a visit at the specified location.  

But @dmdesimone  I'd hope we are pushing people to the data warehouse and this isn't being used so we don't have to put more time into ti?

configuration/reports/reportdescriptors/dataexports/checkins.yml
```diff
@@ -9,6 +9,12 @@ parameters:
   - key: "endDate"
     type: "java.util.Date"
     label: "reporting.parameter.endDate"
+  - key: "visitLocation"
```
mogoodrich, line 12: Note all the java-based reports have a "visiltLocation" parameter instead of a "location" one because that seemed to be the proper name for the parameter in the Java code (to distinguish from encounter location).  In hindsight, we may have wanted to call this visit location in all cases.

configuration/reports/reportdescriptors/dataexports/checkins.yml
```diff
@@ -9,6 +9,13 @@ parameters:
...
+    type: "org.openmrs.Location"
+    label: "reporting.parameter.location"
+    required: false
+    widgetConfiguration:
+      withTag: "Visit Location"
+      defaultIfOnlyOneOption: true
```
mogoodrich, line 18: So for servers with only one visit location, it is automatically selected

configuration/reports/reportdescriptors/dataexports/sql/allpatientswithids.sql
```diff
@@ -30,6 +32,15 @@ select patient_id from patient where voided = 0
 update temp_patients t set t.family_name = person_family_name(patient_id);
 update temp_patients t set t.given_name = person_given_name(patient_id);
 
+-- health center (person attribute)
+update temp_patients t
+left outer join person_attribute pa_hc on pa_hc.person_id = t.patient_id
+	and pa_hc.voided = 0
+	and pa_hc.person_attribute_type_id = (select person_attribute_type_id from person_attribute_type where uuid = '8d87236c-c2cc-11de-8d13-0010c6dffd0f')
```
mseaton, line 39: Dunno how this performs, but might select this into a variable outside this query if it is slow embedded.

configuration/reports/reportdescriptors/dataexports/sql/allpatientswithids.sql
```diff
@@ -30,6 +32,15 @@ select patient_id from patient where voided = 0
...
+left outer join location loc_hc on loc_hc.location_id = pa_hc.value
+set t.health_center_id = pa_hc.value,
```
mseaton, line 41: Shouldn't this be better as `loc_hc.location_id`, since it's actually an integer value?

configuration/reports/reportdescriptors/dataexports/sql/labResultsExport.sql
mseaton: Seems worth fixing.

mseaton (approved):
Made a few comments

configuration/reports/reportdescriptors/dataexports/sql/labResultsExport.sql
mogoodrich: If this will never be used again, I'd rather just remove it--but it is fair that we probably should remove it or fix it.  I'm merging this in as-is, but will add a comment on the ticket about potentially removing entirely.

configuration/reports/reportdescriptors/dataexports/sql/allpatientswithids.sql
mogoodrich: fair, fixed.

configuration/reports/reportdescriptors/dataexports/sql/allpatientswithids.sql
mogoodrich: fair, fixed

PR #740, opened 2026-08-03
HAI-1220: Remove test date restrictions to support historical and future entries

Dutervil:
Updated the test date validation to allow users to select both past and future dates.
Previously, the date picker was restricted by minDate and maxDate, which prevented selecting dates outside the allowed range. The restrictions have been removed to support more flexible date entry. @lnball @louidorjp

### DIFF
--- a/configuration/pih/htmlforms/labResults.xml
+++ b/configuration/pih/htmlforms/labResults.xml
@@ -139,8 +139,6 @@
                 })
                 const encounterDate = getDatePart(new Date(getValue('encounter-date.value'))) || new Date();
                 getField('encounter-date.value').datepicker('option', 'maxDate', new Date());
-                getField('test-date.value').datepicker('option', 'maxDate', null);
-                getField('test-date.value').datepicker('option', 'minDate', null);
                 getField('viral-load.value').change(toggleViralLoadFields);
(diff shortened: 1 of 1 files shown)


lnball (approved):
Looks good @Dutervil .

louidorjp (approved):

louidorjp (changes requested):
You should avoid adding dates in the future.

Dutervil:
` getField('test-date.value').datepicker('option', 'maxDate', null);
                getField('test-date.value').datepicker('option', 'minDate', null);`
These changes will not allow future date

mogoodrich:
Yes, I agree with @louidorjp  that we likely don't want to allow dates in the future... it's possible that this handled by default and we don't need these two fields to be set at all? @Dutervil

Dutervil:
By default, even with this change, it will not allow adding date in the future.  @mogoodrich 
image
 Here is an image as ref

mogoodrich (approved):
Makes sense @Dutervil .  It might be worth checking if we still need these two lines at all, since we are no longer setting the "minDate" and "maxDate" to anything?

louidorjp (approved):

PR #742, opened 2026-08-13
UHM-9402: add new relationships used in Lesotho

chibongho:
See: https://pihemr.atlassian.net/browse/UHM-9261

Closed https://github.com/PIH/lesotho-emr/pull/13 to move these relationships to config-pihemr, in case we need them in other distros in the future.


mseaton (commented):
I don't think @mogoodrich was saying these should be in config-pihemr, was he?  If we aren't using them anywhere else, just leave them in the liberia-content package.  It's easy enough to move over to the pihemr config/content in the future if we want to share elsewhere, right?

mogoodrich:
> I don't think @mogoodrich was saying these should be in config-pihemr, was he? If we aren't using them anywhere else, just leave them in the liberia-content package. It's easy enough to move over to the pihemr config/content in the future if we want to share elsewhere, right?

Yeah, sorry, I wasn't clear in my comment--I was just saying it might be worth seeing if any of these are already in use in other configs, which means it might be time to hoist them to the config-pihemr.

That being said, since all the names are rather generic, I don't have a strong preference as to whether they go in PIH EMR or Lesotho.

chibongho:
Closing this in favor of https://github.com/PIH/lesotho-emr/pull/13

